Issue #7, opened 2018-11-26
No sound after changing output device in Windows


marysiamzawka:
I am using two separate audio outputs in Windows: one for headphones, and another for speakers. I never have both enabled at the same time, but I switch between them a lot.
BambooTracker fails to switch devices correctly.

Steps to reproduce:
- Have only one sound device enabled. Bamboo will produce sound.
- Close Bamboo
- Enable another sound output device and disable the one that was previously used.
- Open Bamboo. Now it will not produce sounds.
- To bring back the sound, open the Configuration window, go to the Sound tab and press OK.

I'm no coder, but I think this could be fixed by making Bamboo default to the, well, "default" (for Windows) sound output.

OPNA2608:
The issue lies here
https://github.com/rerrahkr/BambooTracker/blob/4b675ce7ead78dd98eaf1bb6ee739702054693fd/BambooTracker/gui/mainwindow.cpp#L74-L82

The loaded soundDevice setting goes unchecked, meaning if the device doesn't exist anymore (like when the user disables it), audio initialization fails and `using null output device, none available` gets printed.
Opening the Configuration window will then load the system's default device as the output device and pressing OK will start a new stream with the proper device.
This can be easily simulated by changing the "soundDevice" value in config.json to a non-existent device name and starting the application.

rerrahkr:
Thank @marysiamzawka for your report!

PR #9, opened 2018-11-27
Check if saved soundOutput exists

OPNA2608:
should fix rerrahkr/BambooTracker#7


rerrahkr:
Thanks for your contribution!

PR #10, opened 2018-11-28
Rearrange tab stops for better keyboard navigation

Pigu-A:
Still though I can't seem to rearrange the tab order of those custom widgets that are generated later. Even setTabOrder() isn't working. So the tab stops for FM parameter sliders are still out of place.


rerrahkr:
Thanks for your contribution!

PR #16, opened 2018-11-29
Change envelope reset behavior

Pigu-A:
Also, I think I'll do instrument file importing next since you have the problem understanding some of them.


rerrahkr:

PR #20, opened 2018-11-30
Split FM envelope copy/paste to multiple lines

jimbo1qaz:
Multi-line format and trailing comma allows easily reordering operators in a text editor like Sublime or Notepad++.

I'm already using this change, it makes algorithm switching (and reordering ops) very easy.

This code could use some unit tests. And the massive regex could be replaced.


rerrahkr:

PR #30, opened 2018-12-16
Create BambooTracker.1

alexmyczko:
a basic first manual page, feel free to enhance as needed.

### DIFF
--- a/BambooTracker.1
+++ b/BambooTracker.1
@@ -1,0 +1,25 @@
+.TH BambooTracker 1 "December 15 2018"
+.SH NAME
+BambooTracker \- YM2608 (OPNA) tracker
+.SH SYNOPSIS
+.B BambooTracker
+.RI " files" ...
+.br
+.SH DESCRIPTION
+This manual page documents briefly the
+.B BambooTracker
+command.
+.PP
+\fBBambooTracker\fP is a tracker for YM2608 (OPNA) which was used in NEC PC-8801/9801 series computers.
+.SH OPTIONS
+No options are supported.
+.SH SEE ALSO
+.BR cheesecutter (1),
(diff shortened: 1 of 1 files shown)


BambooTracker.1
```diff
@@ -0,0 +1,25 @@
+.TH BambooTracker 1 "December 15 2018"
+.SH NAME
+BambooTracker \- YM2608 (OPNA) tracker
+.SH SYNOPSIS
+.B BambooTracker
+.RI " files" ...
+.br
+.SH DESCRIPTION
+This manual page documents briefly the
```
OPNA2608:
```suggestion
This manual page briefly documents the
```
or
```suggestion
This manual page quickly summarizes the
```
would be better / more correct.

OPNA2608:
I think it'd make sense to document all/most of the key commands described in the [README](https://github.com/rerrahkr/BambooTracker/blob/86428063251b90a859633b1e8adf5da1215f2a8c/README.md), especially since there isn't much else to say in a man page about a switchless GUI application imo.

alexmyczko:
I agree, however to have a debian package, it is required to have one:
https://www.debian.org/doc/debian-policy/ch-docs.html
although many don't (it is considered a bug):
https://lintian.debian.org/tags/binary-without-manpage.html

rerrahkr:
I really do appreciate everyone's help as I do not know much about Linux.

BambooTracker.1
rerrahkr: I will fix it when I edit other documents.

rerrahkr: fixed 7a934324d9e569d2e62200cad954dda2c148f4fd

Issue #17, opened 2018-11-30
Enhancement: use system's config directory


OPNA2608:
While the current system of storing the configuration `config.json` in the current directory works well enough, it would be desirable to save it to and load it from a fixed directory in a cross-platform way (meaning no hardcoded Windows path or otherwise Windows-only solution).

Using QSettings in IniFormat (or NativeFormat, which uses the Registry on Windows) seems convenient for this purpose:
http://doc.qt.io/qt-5/qsettings.html

This isn't anything super urgent, so it doesn't require to be implemented immediately, but it'd be nice to see in the future _:)_

OPNA2608:
Implemented with https://github.com/rerrahkr/BambooTracker/pull/31, closing!

PR #31, opened 2018-12-16
Replacing Json config with QSettings

OPNA2608:
This addresses #17.

This PR replaces the Json-based configuration stored in `./config.json` with a pwd-independent `QSettings` setup and removes the now unused `Json` class.
The structure of the new `ConfigurationHandler` is about the same as the `Json` class, but replaces the Json-construction and file reading with `QSettings` calls.

Under *nix, the configuration can now be located under:
  `~/.config/BambooTracker/BambooTracker.ini`
[It appears this file will be somewhere under APPDATA on Windows](http://doc.qt.io/qt-5/qsettings.html#setPath), but I can't check what the exact location is.

The `ConfigurationHandler::organization` and `ConfigurationHandler::application` fields can be adapted as you see fit, they'll primarily adjust the path of the resulting .ini file, as explained on the doc page linked above. (`~/.config/<organization>/<application>.ini` under *nix)

---
**After merging this, BambooTracker will use the same configuration, no matter where you start it from, but you should mention in the next release notes that users will have to readjust their settings!**


OPNA2608:
I added another commit that reduces the `weak_ptr.lock()`ing and reads the passed config during the `readConfiguration` method and writes the currently set value back in case a field is missing from the configuration file or the whole file doesn't exist.
* `std::__cxx11::string` have to be capsuled in `QString`s because they don't natively convert to the required `QVariant` type
* `size_t` fields need to be handled like this, else compilation fails because `QVariant(long unsigned int)` couldn't be handled:
```cpp
QVariant pageJumpLengthWorkaround;
pageJumpLengthWorkaround.setValue(configLocked->getPageJumpLength());
```

@rerrahkr Does this look okay or do you suggest changes?

OPNA2608:
Ah whoops, I forgot about that :man_facepalming:. Thanks for pointing it out!

rerrahkr:
Thanks for your contribution!

PR #36, opened 2018-12-18
Update BambooTracker.desktop

trebmuh:
adds French comment and Generic name, and remove the final point (consistency with others desktop files)

### DIFF
--- a/BambooTracker.desktop
+++ b/BambooTracker.desktop
@@ -2,9 +2,11 @@
 Name=BambooTracker
 GenericName=OPNA Tracker
 GenericName[jp]=OPNAのトラッカーです
-Comment=A tracker for the YM2608 (OPNA), which was used in NEC PC-8801/9801 series of computers.
+GenericName=Tracker OPNA
+Comment=A tracker for the YM2608 (OPNA), which was used in NEC PC-8801/9801 series of computers
 Comment[jp]=このアプリケーションはNEC PC-8801/9801シリーズに搭載されていたFM音源YM2608(OPNA)向けのトラッカーです。
-Comment[de]=Ein Tracker für den YM2608 (OPNA), welcher in NECs PC-8801/9801 Serien an Computern verwendet wurde.
+Comment[de]=Ein Tracker für den YM2608 (OPNA), welcher in NECs PC-8801/9801 Serien an Computern verwendet wurde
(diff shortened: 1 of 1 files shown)


BambooTracker.desktop
```diff
@@ -2,9 +2,11 @@
 Name=BambooTracker
 GenericName=OPNA Tracker
 GenericName[jp]=OPNAのトラッカーです
-Comment=A tracker for the YM2608 (OPNA), which was used in NEC PC-8801/9801 series of computers.
+GenericName=Tracker OPNA
```
OPNA2608: I think you forgot the locale specification
```suggestion
 GenericName[fr]=Tracker OPNA
```

BambooTracker.desktop
```diff
@@ -2,9 +2,11 @@
...
+Comment=A tracker for the YM2608 (OPNA), which was used in NEC PC-8801/9801 series of computers
 Comment[jp]=このアプリケーションはNEC PC-8801/9801シリーズに搭載されていたFM音源YM2608(OPNA)向けのトラッカーです。
-Comment[de]=Ein Tracker für den YM2608 (OPNA), welcher in NECs PC-8801/9801 Serien an Computern verwendet wurde.
+Comment[de]=Ein Tracker für den YM2608 (OPNA), welcher in NECs PC-8801/9801 Serien an Computern verwendet wurde
+Comment[de]=Un tracker pour le YM2608 (OPNA), qui fut utilisé dans la séries d'ordinateurs NECs PC-8801/9801
```
OPNA2608:
```suggestion
Comment[fr]=Un tracker pour le YM2608 (OPNA), qui fut utilisé dans la séries d'ordinateurs NECs PC-8801/9801
```

OPNA2608:
Thanks for catching the extra dots, I only run a bare WM which doesn't utilize .desktop files!

trebmuh:
Good catch @OPNA2608 :+1: 

That's what happen when you:
- try to package another software
- have a discussion about a different topic on IRC
- manage to fix a coat rack without having propers tools to use
- try to answer all and every questions of a 8yo niece
- think you should focus on only one thing at a time
- and because that's not enough, decide to contribute a 2-liners to BambooTracker

:smile:

rerrahkr:

PR #37, opened 2018-12-21
Support the WOPN instrument format

jpcima:
Hello. It's a first attempt to add @Wohlstand's OPNI file format. A [specification](https://github.com/Wohlstand/OPN2BankEditor/blob/master/Specifications/WOPN-and-OPNI-Specification.txt) is available here.

The specification is for two formats : a `.wopn` file is a bank, `.opni` is a single instrument.
This patch adds `.opni` only, using a MIT-licensed part from libOPNMIDI.
(This same library code permits to add `.wopn` support later easily; it should be only a matter of implementing user interface where patches can be imported from a list)

For information, WOPN is the format of [libOPNMIDI](https://github.com/Wohlstand/libOPNMIDI) and [OPN2BankEditor](https://github.com/Wohlstand/OPN2BankEditor), it has some adoption in a few projects, among which is the engine gzdoom, and a FM synth which I make. The bank [DMXOPN2](https://github.com/papiezak/DMXOPN2) is current work using WOPN.

For single instrument files, it's possible to have a few here: (may have a few outdated or bad)
https://github.com/Wohlstand/OPN2BankEditor/tree/master/Bank_Examples/Instruments

### DIFF
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -161,7 +161,8 @@ SOURCES += \
     gui/color_palette.cpp \
     gui/command/pattern/paste_overwrite_copied_data_to_pattern_qt_command.cpp \
     command/pattern/paste_overwrite_copied_data_to_pattern_command.cpp \
-    io/file_io_error.cpp
+    io/file_io_error.cpp \
+    format/wopn_file.c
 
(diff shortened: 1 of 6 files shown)


rerrahkr:
Thanks for your contribution!

PR #38, opened 2018-12-22
Support importing instruments from banks in WOPN format

jpcima:
This follows WOPN instrument support, adding the support of bank files.
In the GUI, I added `Import From Bank File`, which permits to load a choice of instruments from a `.wopn` file.
To do this, I added a dialog and I introduced a model for banks which I thought to be appropriate.
The instruments are picked from a list widget and it's possible to load several of them at once.

EDIT: add minor fix where the error did not indicate the good file type

### DIFF
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -162,7 +162,9 @@ SOURCES += \
     gui/command/pattern/paste_overwrite_copied_data_to_pattern_qt_command.cpp \
     command/pattern/paste_overwrite_copied_data_to_pattern_command.cpp \
     io/file_io_error.cpp \
-    format/wopn_file.c
+    format/wopn_file.c \
+    instrument/bank.cpp \
+    gui/instrument_selection_dialog.cpp
 
 HEADERS += \
     gui/mainwindow.hpp \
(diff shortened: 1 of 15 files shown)


rerrahkr:
Thanks for your contribution!

jpcima:
Great. :+1: Thanks for accepting the format. I wish to ask one thing:

The instrument structure contains `note offset` data. It's in semitone units.
Usually the instrument maker will set this to pitch up or down by some octaves.

Is there a way to make instruments use this information? I've had some difficulty finding how to accomplish this in a good way.

rerrahkr:
I think it is able to set WOPN's note offset data in FM arpeggio sequence by selecting absolute type and entering offset for 1 step.

jpcima:
I implement the suggestion as https://github.com/jpcima/BambooTracker/commit/d32c3b061b404edaf8ac7ce8ba6689e750bca74d.

It's working but has one issue: when it has a -12 semitone offset for example, playing the instrument on the octave 0 hits a lower limit on the note number, and these pitches are not playable.

In OPN2BankEditor, the emulator MAME YM2608 is able to play these frequencies.
There is for example the melodic patch 17 `Percussive Organ` from [xg.wopn](https://github.com/Wohlstand/OPN2BankEditor/blob/master/Bank_Examples/xg.wopn) having -12 offset.

rerrahkr:
Since the tracker holds the pitch data in the table, it can not refer to the pitch outside the range.
And as mentioned in #40, the range of FM pitch will be 1 octave lower.

rerrahkr:
I added the suggenstion in 31d8bf61affdf93de10224d34e2394abbfedc03f.
Thank you!

PR #39, opened 2018-12-22
Fix invalid memory access on main window destruction

jpcima:
Solves bad access into freed memory when the program is closed, found by address sanitizer.
On destruction of the comstack, it triggers a slot on the main window whose `ui` has been freed prior.

To prevent the problem, I ensure to destroy the `ui` last.
I put the variable in RAII and it's working because the members destroy in reverse order.


rerrahkr:
Thanks for your contribution!

PR #52, opened 2019-01-04
Add instructions for FreeBSD users

ehaupt:
Now that there is an official [FreeBSD port](https://svnweb.freebsd.org/ports/head/audio/bambootracker/) for bambootracker. Add some instructions.


rerrahkr:

PR #64, opened 2019-02-13
Revert and actually fix the envlope reset behavior

Pigu-A:
Apparently I did the older fix wrong and made key off happens one tick too early, which somebody really didn't like it...

Doing note off in every slots immediately before note on will allow the envelope to retrigger with the last level value retained. This is a correct behavior from Deflemask and various MML drivers.


rerrahkr:
Thanks for your contribution!

Issue #83, opened 2019-03-25
Cannot build in Qt Creator (.ts .qm translations)


jimbo1qaz:
I'm trying to build BambooTracker master, from Qt Creator 4.5.2, Based on Qt 5.9.5 (GCC 7.3.0, 64 bit).

Out-of-path build is enabled.

When `make` runs, it encounters error:

```
:-1: error: No rule to make target 'res/lang/bamboo_tracker_fr.ts', needed by '.qm/bamboo_tracker_fr.qm'.  Stop.
```

Commenting out the following lines fixes the issue:
```
TRANSLATIONS += \
    res/lang/bamboo_tracker_fr.ts \
    res/lang/bamboo_tracker_ja.ts
```

res/lang/bamboo_tracker_fr.ts and res/lang/bamboo_tracker_ja.ts both exist.

OPNA2608:
I didn't test building the workaround way with QtCreater back then, mainly because I prefer vi and my main goal was only to make it compile with qmake and make so I can package it.

That aside, both Debug and Release build works fine for me. I deleted .qmake.stash, Makefile and .qm/ before building so it'd have to regenerate the translation files.
QtCreater 4.5.2, Qt 5.9.5, GCC 7.3.0, 64 bit) as well.
![grafik](https://user-images.githubusercontent.com/23431373/54898488-d546e980-4ecc-11e9-8b7d-a4d26a54ec95.png)

PR #85, opened 2019-03-25
fix Out-of-Path build for Qt version workaround

OPNA2608:
( closes https://github.com/rerrahkr/BambooTracker/issues/83 )

By prefixing the .ts files with `$$PWD`, we will point to the absolute path of the files when building into another directory.


rerrahkr:
Thanks for your contribution!

PR #89, opened 2019-03-28
Draft: More dynamic keyboard layout mapping

### LINKED ISSUE #70: Permit piano keyboard keys to be rebindable
The keybinds for adding notes, as the README mentions, are:
> There are two rows of a piano keyboard:
> ```
> Current octave+1
>  2 3  5 6 7  9
> Q W ER T Y UI O
> 
> Current octave
>  S D  G H J  L
> Z X CV B N M, .
> ```

This works fine on a QWERTY keyboard.

Using a non-QWERTY however, like the [German QWERTZ](https://en.wikipedia.org/wiki/QWERTZ), [French AZERTY](https://en.wikipedia.org/wiki/AZERTY) or, well, the [Russian JCUKEN using Cyrillic script](https://en.wikipedia.org/wiki/JCUKEN) can either take some getting used to or might just not work at all. (I can only speak for myself, using a QWERTZ keyboard.)

I propose that, at least, these keys should be rebindable. They can stay the default, I highly doubt that the majority of users will be using a non-QWERTY layout, but the user should be able to change the mapping and save it how they see fit. The groupbox for changing the keys to change the current octave etc. could be moved into the same settings window tab as well.

Perhaps offering some presets for different layouts as well would be the icing on the cake.

OPNA2608:
Closes #70 

~~This is a draft PR. It's not yet feature-complete~~ This PR is now feature-complete, but some feedback and tips on the design and how to structure everything would be appreciated, as I'm still very much a beginner at C++ and Qt!

### The goal:
- Add a customizable keyboard layout option.
- Improve the layout handling structure under the surface.

### Current state:
- `std::map`s for all currently supported layouts as well as a custom one have been added.
- A new tab for key mapping has been added to the configuration dialog.
- The layout selector and Special Keys GroupBox have been moved to said key mapping tab.
- A custom layout section has been designed and completed function-wise.
- The custom layout can be saved, loaded and reset (to QWERTY).

### Problems / Feedback points:
- I'm not a designer, but I tried to have the key mapping tab look nice.
- There's still abit of code duplication between `mainwindow.cpp` and `pattern_editor_panel.cpp`.
I think this needs to be moved into a separate class with a method declaration like this?:
```c++
static JamKey LayoutHandler::getJamKeyFromLayoutMapping(Qt::Key key);
```

Let me know what you think so far!


OPNA2608:
Interesting that AppVeyor's compilation fails, as I can compile everything just fine on my end?

Last attempted compile command from AppVeyor vs same command from my local build.
https://ci.appveyor.com/project/rerrahkr/bambootracker/build/job/x4m7rxu14674354k#L64
```
g++ -c -pipe -O2 -std=gnu++1y -Wall -W -D_REENTRANT -fPIC -DQT_DEPRECATED_WARNINGS -D__LINUX_ALSA__ -DQT
_NO_DEBUG -DQT_MULTIMEDIA_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -I. -Ichips -
Istream -Iinstrument -Icommand -Imodule -Iio -isystem /usr/include/x86_64-linux-gnu/qt5 -isystem /usr/in
clude/x86_64-linux-gnu/qt5/QtMultimedia -isystem /usr/include/x86_64-linux-gnu/qt5/QtWidgets -isystem /u
sr/include/x86_64-linux-gnu/qt5/QtGui -isystem /usr/include/x86_64-linux-gnu/qt5/QtNetwork -isystem /usr
/include/x86_64-linux-gnu/qt5/QtCore -I. -isystem /usr/include/libdrm -I. -I/usr/lib/x86_64-linux-gnu/qt
5/mkspecs/linux-g++ -o bamboo_tracker.o bamboo_tracker.cpp
```

jpcima:
It's looking promising, I would suggest a few things.

- indentation style is mismatched with original code, because original has tabs
- fixed mappings should be static and const + moved to implementation side
- I think you should not include the header `<Qt>`
- the selected tab should be the first one on opening the dialog, currently it's the last
- saving/loading key settings
- avoid including `jam_manager` from the header, if possible

OPNA2608:
> - indentation style is mismatched with original code, because original has tabs

QtCreator insists on replacing all tabs of lines I touch with spaces, I gotta figure out how to turn that off for a project. I noticed that after committing, unfortunately.

> - fixed mappings should be static and const + moved to implementation side
> - the selected tab should be the first one on opening the dialog, currently it's the last
> - saving/loading key settings

:+1: 

> - I think you should not include the header `<Qt>`
> - avoid including `jam_manager` from the header, if possible

They were needed for the `Qt::Key` & `Qt::Key_...` and `JamKey` types, respectively. I could use raw `int` instead I think, but I'd have to cast them to the respective types/enums for some method calls anyway. Am I missing something to avoid importing them besides manually declaring them myself?

jpcima:
You are surely right about `<Qt>` and the `Key` enum.

As `jam_manager` is concerned, I don't think you need this in a header when the map definitions are moved to .cpp side. A forward definition would be sufficient. (eg `enum class JamKey;`)

OPNA2608:
@jpcima 

- Indention should be fixed now as far as I can tell. I kept the space indention for vertical alignment between the `Qt::Key`s and `JamKey`s for now, but I can replace / drop them later as well.
- I moved all the map definitions to `configuration.cpp`, the `jam_manager.hpp` import from the `.hpp` to the `.cpp` and added a forward definition for `JamKey` in the `.hpp`.
- Fixed the default tab in the configuration dialog not being the General tab.

I plan to add the missing keys in the custom layout section, the wiring between the custom settings and settings map and the saving/loading of the key settings over the next few days whenever I have time.

rerrahkr:
As I avoid using Qt as much as possible except for gui, it would be better to hold keys in utf8 `std::string` as well as special keys in configuration.
When checking pressed key, it can find `JamKey` from map by `QKeySequence` and `std::find` like:
```c++
auto it = std::find_if(selectedLayoutMapping.begin(), selectedLayoutMapping.end(),
				 [key](const std::pair<std::string, JamKey>& t) -> bool {
				 return (QKeySequence(key).matches(QKeySequence(t.first)) == QKeySequence::ExactMatch);
});
if (it != selectedLayoutMapping.end()) {
	return it.second;
}
else {
	throw std::invalid_argument("Unmapped key");
}
```

> Interesting that AppVeyor's compilation fails, as I can compile everything just fine on my end?

Compiling works fine in my environment, so I don't know why Appveyor causes errors...

OPNA2608:
@rerrahkr 
>As I avoid using Qt as much as possible except for gui

That sounds like a good idea. I fixed the offset from the current octave being calculated wrong and replaced the `Qt::Key`s with u8 `std::string`s as you suggested today. Your code needed 2 changed to work though:
```diff
-		return (QKeySequence(key).matches(QKeySequence(t.first)) == QKeySequence::ExactMatch);
+		return (QKeySequence(key).matches(QKeySequence(QString(t.first))) == QKeySequence::ExactMatch);
```
```diff
-	return it.second;
+	return (*it).second;
```

rerrahkr:
> Your code needed 2 changed to work though

Thank you for telling me the fix!

OPNA2608:
I implemented custom layout saving and loading today. With that, the main goal of the PR has been reached and it can be reviewed.

I think it would be better to move everything into a separate LayoutHandler class instead of defining and redefining these mapping and small helper maps all over the GUI, I'll consider doing that when I have more time again. :slightly_frowning_face:

OPNA2608:
I resolved rebased the PR on the current HEAD and hopefully solved all merge conflicts. Let's see if AppVeyor can build it now :slightly_smiling_face:

BambooTracker/gui/configuration_dialog.cpp
```diff
@@ -24,57 +25,98 @@ ConfigurationDialog::ConfigurationDialog(std::weak_ptr<Configuration> config, QW
 		emit applyPressed();
 	});
 
+	std::shared_ptr<Configuration> configLocked = config.lock();
```
OPNA2608: I mass-replaced all the separate `config.lock()`s and implicit unlocks with a single locked `configLocked`. Let me know if you'd rather not have that, if I should move it into a separate commit/PR, …

rerrahkr: It's OK to push the changes to this PR.

rerrahkr:
It works well, thanks for your contribution!

PR #90, opened 2019-03-28
Indicate the correct version in specification documents

jpcima:
It was forgotten to change the document versions after updating.


rerrahkr:
Thank you for correcting version numbers!

Issue #93, opened 2019-04-04
Improve SSG-EG emulation


freq-mod:
Kinda similar to #77 - it may be not the OPNA emulation problem, but anyway...
SSG-EG sounds weird - totally different than on YM2612 trackers/utilities, which is weird, since YM2612 is based directly on YM2608's FM block and there should be no difference in SSG-EG timbre...

https://instaud.io/3vIA - SSG-EG instrument recorded using BambooTracker
https://instaud.io/3vIC - the same instrument, recorded using OPN2 Bank Editor, with "Nuked OPN2" emulator, one of the most accurate soundchip emulators ever

The difference between two recordings are huge - 2608's SSG-EG sounds **very** weak compared to 2612. 
[ssgeg.zip](https://github.com/rerrahkr/BambooTracker/files/3045354/ssgeg.zip) - zip file with instrument and emu recordings in question

rerrahkr:
Thanks for your report. I checked the instrument on real chip (BambooTracker -> SCCI -> c86ctl ->G.I.M.I.C).
[real.zip](https://github.com/rerrahkr/BambooTracker/files/3048122/real.zip)
This sound is similar to Nuked OPN2, so it seems that MAME emulation is wrong.

papiezak:
Well, that interesting.
I recall @jpcima tried to correct SSG-EG emulation on Neko Project 2 OPNA emulation core using Nuked OPN2 code, the result was a bizarre cross between two 😆  Maybe with some tweaks, MAME can be improved.
Thanks for checking on a real chip anyway.

jpcima:
>This sound is similar to Nuked OPN2, so it seems that MAME emulation is wrong.

Interesting, then it means Nuke.YKT had the correct intuition on this.

First thing, I'll be interested to compare MAME's independent YM2612 core, vs MAME generic FM code (which generates various YM* chips, one being YM2608). In our experience, these two are generating the SSG-EG differently.

Also it could be examined how much work it takes to bolt some features onto Nuked-OPN2, to make it into a semi-accurate OPNA-ish that Bamboo Tracker can use.

>I recall @jpcima tried to correct SSG-EG emulation on Neko Project 2 OPNA emulation core using Nuked OPN2 code, the result was a bizarre cross between two

It's not a really accurate statement, the SSG-EG work introduced was not Nuked's but mine.
I just had the EG curve match, in approximation, what Nuked EG would output.
[...]

PR #94, opened 2019-04-06
Add Nuked OPN-Mod emulator

jpcima:
#93

The change allows to use very high-fidelity FM core reverse-engineered by Nuke.YKT.

Over the basic FM core, which is OPN2, I introduced the useful feature set of OPNA.
A part of rhythm channel handling is a backport of MAME code.
This means this modified core has ability to play all OPNA channel kinds.
For clarity, I have maintained a commit log as clean I can.

The configuration dialog contains a choice setting for the emulator.

**Note:** one thing not yet made: supposedly, OPN2 clips channels at lower level than OPNA, and the core needs a change to reflect this difference.
@papiezak are you able to produce a module file which showcase the clipping problem?


papiezak:
>@papiezak are you able to produce a module file which showcase the clipping problem?

[clip.zip](https://github.com/rerrahkr/BambooTracker/files/3051363/clip.zip) - there it is. OPN2(C) apperently uses 9-bit accumulator per channel and therefore any waveform gets clipped if it's loud enough. In YM2608 case, as @nukeykt explains: 
>Checked YM2610 die shot. At first it drops LSB bit of operator output(from 14-bit to 13-bit) and then sums them into 18-bit global accumulators for left and right channels. Then these accummulated values gets clamped to 16-bits. I'm pretty much sure that YM2608 does the same.

jpcima:
Thanks, I added more bits, and allowed quantization to be settable in the source.
I don't know if it's the right amount but it fixed the case at least.

Hopefully this completes the work, and there can be fair Nuked vs Real chip comparison basis.

papiezak:
Now it doesn't clip, so that thing is done. Thanks!

papiezak:
Aren't there some leftovers related to YM2612 Channel 6 PCM mode?  If yes, can they be removed, since it's irrelevant to YM2608 emulation?

jpcima:
Yes, I haven't bother to remove DAC but you're right, this code can go.

Another thing is the mute flags of the extra interfacing code from vgmplay.
They're not needed for normal use, and actually I don't understand the logic of them at all. I'm not convinced they even work. In libopnmidi context we don't use it either.

jpcima:
I made the mistake in drum key-off not to check the channel bits in fact, so I fixed it.
When tried, I cannot reproduce this sound on drum channels 5 and 6; I'll plot this again and search for the circumstance which creates the tone difference. (although, I have a little idea why this may be)

jpcima:
I identified a problem, fixed it, revised the ADPCM-A commit.
I also checked mixing to match the volumes more accurately, and verified on scope capture.

For the last significant feature missing, being ADPCM-B/YMDELTA-T, it's still not implemented.
From what I can tell, it is easily done by plugging in the source file "ymdeltat" from MAME; but it will be more comfortable to implement when BambooTracker will have the support.

rerrahkr:
I confirmed drum's bugs were fixed, thanks!

> it will be more comfortable to implement when BambooTracker will have the support.

I agree. First of all, I think it is necessary to make ADPCM available when using MAME.
I will merge this PR.

jpcima:
Please do. :+1: 
I don't need to add other things.

rerrahkr:
Thanks for contributions!

rerrahkr:
@jpcima Now I'm working on ADPCM feature in `ADPCM` branch, and at aa120023e608b40e596447204a8333c88b817691 ADPCM jamming was added, worked fine on MAME. It still not supports some features such as note entry, but I implemented operations to change the register. Would you mind implementing ADPCM in Nuked-OPN Mod?

jpcima:
Yes, no promise, I will try to complete this part fast.

jpcima:
There exists a buffering issue which breaks integrity of the register stream..

Nuked needs to run some delays in between register writes, it needs an intermediate buffer that processes this.
May it be the case that sound generation is called in a different thread, at the same time that dummy data is fed into ADPCM?
When I checked register behavior in a context outside of BambooTracker with the same register input, there wasn't a buffering problem.

rerrahkr:
There is possibility to generate samples in the audio stream during dummy feeding in latest commit. But the tracker will implement to stop register writing and audio streaming during sample feed (feeling necessary from my experience playing songs using ADPCM on real chips).
I need to modify the code to stop audio streaming during feeding dummy data before merging your new Nuked.

jpcima:
I should have mentioned, the WIP is on my Nuked-ADPCM branch derived from yours.
I expect it should work better after fixing the register stream.
It's where I last stopped since I'm also busy now with some other things. I'll dig more into the matter later.

rerrahkr:
I changed jam action at 0df263a60b2ef5f7811e0ddc4c600943ab47087e. Dummy sample is set to register by adding a new ADPCM instrument. During sending sample to the register, the audio stream is paused.

jpcima:
Based on 0df263a, there is still a register writing anomaly.
When registers are logged in MAME vs Nuked's `OPN2_Write` entry, it's observed there is a sequence of writes which doesn't reach it.

So it's needed that I investigate the buffering, especially when they are long bursts of messages arriving. The role of the buffer is to ensure that register entry be at least separated by the minimum chip cycles that it needs between these writes.

jpcima:
@rerrahkr. Hi. still working on commit 0df263a,
I observe `OPN2_Reset` calls arriving in the middle of ADPCM registers being written.
It's this which is the origin of buffering trouble.

Have there been progress regarding sample loading?

rerrahkr:
The timing of a loading dummy sample is changed at e5b255a8d2eb46af7972de5bcd460769a80c7775. When pressing "Import" button in Waveform of ADPCM instrument editor, data is write to the register. Currently I'm working on sample import from .wav file.

PR #95, opened 2019-04-07
Fix MML format edit alignment

### LINKED ISSUE #87: SIGSEGV with buggy FM envelope input editor, value mapping display misalignment
I've noticed two bugs within the Configuration->Input tab:

---

#### 1) Name -> Specification association
While renaming a new or any of the existing FM envelope format specifications, if the currently entered name is equal to an existing specification's name, the currently edited specification as well as the one it shared its name with for any amount of time will be associated with each other: They share the same editable list of text -> FM operator mappings and deleting one of them will also delete the other.

Example: I want to add an alternate specification of the PMD format [as described here](https://pigu-a.github.io/pmddocs/pmdmml.htm#3-1) and call it PMD-DT2. (because it adds a secondary Detune value to each operator specification)

1) Have a FM envelope text specification called "PMD".
2) Create a new FM envelope text specification.
3) Edit the name from "Set …" to "PMD-DT2".

While typing, as soon as the textbox contains the string "PMD", PMD-DT2 will be linked to PMD.

- If you try to edit PMD-DT2, it'll have the same content as PMD. A SIGSEGV is achievable here by changing a value, then pressing OK and selecting PMD. (valgrind trace below)
- If you save, PMD-DT2 will replace PMD.
- If you delete PMD-DT2, PMD will be deleted as well. (and vice versa)

[Here is the valgrind log. It has alot of noise at the start, the SIGSEGV happens at line 1064.](https://github.com/rerrahkr/BambooTracker/files/3008358/BambooTracker.log)

#### 2) Bugged mapping display
When adding a value while value number 0 or no value (when pressing Edit followed by Add without highlighting any existing value) is selected, the values will be misaligned. This is only a temporary visual bug, as closing the window and pressing Edit again will fix this.

grafik

rerrahkr:
Fix #87.

On windows 10, 4b6e56b fixes the type alignment but it seems not to fix it on Linux.
In this commit, I add code to insert and remove a dummy row at the 2nd row when some change is performed on the 1st row. (I also fix to update digit type numbers when remove a row.)
@OPNA2608 could you check if this prevents misalignment on Linux?

### DIFF
--- a/BambooTracker/gui/fm_envelope_set_edit_dialog.cpp
+++ b/BambooTracker/gui/fm_envelope_set_edit_dialog.cpp
@@ -54,6 +54,8 @@ void FMEnvelopeSetEditDialog::swapset(int aboveRow, int belowRow)
 		tree->setItemWidget(below, 1, belowBox);
 	}
 
+	if (!aboveRow || !belowRow) alignTreeOn1stItemChanged();	// Dummy set and delete to align
+
 	for (int i = aboveRow; i < ui->treeWidget->topLevelItemCount(); ++i) {
 		ui->treeWidget->topLevelItem(i)->setText(0, QString::number(i));
 	}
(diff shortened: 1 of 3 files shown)


OPNA2608:
That appears to have fixed the misalignment on Linux as well, :+1:!

rerrahkr:
Thanks for checking!

PR #97, opened 2019-04-11
Fix Y12 instrument loader

jpcima:
Some fixes to Y12 loader.
- operator order 1-3-2-4 (was 1-2-3-4)
- mask the unused bits
- adapt to Bamboo tracker the register value of SSG-EG


rerrahkr:
Thanks for your contribution!

Issue #98, opened 2019-04-11
YM2203 and YM2419 vgm/s98 export


pedipanol:
I'm aware the purpose of the tracker is to solely to emulate the OPNA soundchip. But I think it would be nice to have the ability to export vgm/s98 files in this format.

From what I heard the register writes are basically the same throughout these soundchips, since the OPNA is a derivative of YM2203 and the SSG is a variant of the YM2419, so I imagine it wouldn't be difficult.

I know djtuBIG-MaliceX made a SSG to AY converter, and I believe this is the source code but I don't know if it would be useful https://github.com/djtuBIG-MaliceX/my-random-tools/tree/master/vgm-file-related/vgm_ayconv

The purpose of me requesting this is just so songs that only use the YM2203 channels or just SSG are properly saved in a format that doesn't use other channels. Ideally it would be if Bambootracker would allow to switch soundchips on the module itself but I think that's too much to ask all things considered.

jpcima:
It's an easy conversion indeed, and also I have been interested in having such a feature myself.
Lately I did work some work in this direction, outside of BambooTracker and based on vgm; it will be practical to add such a thing in the exporter directly.

I thought of adding a set of export options.
- FM to OPNA/OPN2/OPN (OPM?)
- SSG to internal or external YM2149

The YM2203, specifically, should have clock at 3993600 to preserve the timbres.
(has clock divisor of 72, as observed from MAME)

cyberic99:
Yeah, OPM with 8 FM channels would be awesome!

pedipanol:
@cyberic99 Implementing OPM into the S98/VGM export would already require more attention (which is why @jpcima added a ? on the option), while OPN soundchips and the YM2419 being from the same family are almost directly compatible between each other, so the only potential problem on the export would be the garbage additional data and the clock setting.

The OPM soundchip operates in a different way (additional channels, additional instrument data, no SSG-EG, no FM3Extend, noise mode on channel 8, different hardware LFO settings, etc.), so implementing it into the tracker would require a totally new sound emulation core. Unlike my suggestion of stripping down OPNA exclusive channels, since that would still use the OPNA emulation core.

Even if it would be viable to add it I don't think it should be done yet, because I think it would be better for the tracker to perfect the use of the OPNA before trying a totally different soundchip such as the OPM.

cyberic99:
@pedipanol thanks for your explanations.
I didn't know about the noise mode.

[...]

PR #99, opened 2019-04-11
VGM export for various chip targets

jpcima:
#98
 
Add to VGM exporter an ability to output for a different chip, or separate AY chip.
Dialog is updated and displays feature availability for the target.

- FM : OPNA, OPN2, OPN
- SSG : Internal, AY8910, YM2149

YM2203 can use internal SSG but will be at different clock setting.

**Note:** I should really clean up the garbage register writes for chip other than OPNA

**Edit1:** have mixed up AY8910 type & flag fields, fixed
**Edit2:** rebased on master, eliminated conflict

### DIFF
--- a/BambooTracker/bamboo_tracker.cpp
+++ b/BambooTracker/bamboo_tracker.cpp
@@ -957,7 +957,7 @@ bool BambooTracker::exportToWav(std::string file, int loopCnt, std::function<boo
 	}
 }
 
-bool BambooTracker::exportToVgm(std::string file, bool gd3TagEnabled, GD3Tag tag, std::function<bool()> f)
+bool BambooTracker::exportToVgm(std::string file, int target, bool gd3TagEnabled, GD3Tag tag, std::function<bool()> f)
 {
 	int tmpRate = opnaCtrl_->getRate();
 	opnaCtrl_->setRate(44100);
@@ -974,7 +974,7 @@ bool BambooTracker::exportToVgm(std::string file, bool gd3TagEnabled, GD3Tag tag
 	uint32_t loopPoint = 0;
 	uint32_t loopPointSamples = 0;
 	std::shared_ptr<chip::VgmExportContainer> exCntr
-			= std::make_shared<chip::VgmExportContainer>(mod_->getTickFrequency());
+			= std::make_shared<chip::VgmExportContainer>(target, mod_->getTickFrequency());
 	opnaCtrl_->setExportContainer(exCntr);
 	startPlayFromStart();
 	exCntr->forceMoveLoopPoint();
@@ -1004,7 +1004,7 @@ bool BambooTracker::exportToVgm(std::string file, bool gd3TagEnabled, GD3Tag tag
 	opnaCtrl_->setRate(tmpRate);
 
 	try {
-		ExportHandler::writeVgm(file, exCntr->getData(), CHIP_CLOCK, mod_->getTickFrequency(),
+		ExportHandler::writeVgm(file, target, exCntr->getData(), CHIP_CLOCK, mod_->getTickFrequency(),
 								loopFlag, loopPoint, exCntr->getSampleLength() - loopPointSamples,
 								exCntr->getSampleLength(), gd3TagEnabled, tag);
 		f();
--- a/BambooTracker/bamboo_tracker.hpp
+++ b/BambooTracker/bamboo_tracker.hpp
@@ -186,7 +186,7 @@ public:
 
 	// Export
 	bool exportToWav(std::string file, int loopCnt, std::function<bool()> f);
-	bool exportToVgm(std::string file, bool gd3TagEnabled, GD3Tag tag, std::function<bool()> f);
+	bool exportToVgm(std::string file, int target, bool gd3TagEnabled, GD3Tag tag, std::function<bool()> f);
 	bool exportToS98(std::string file, bool tagEnabled, S98Tag tag, std::function<bool()> f);
 
 	// Backup
--- a/BambooTracker/chips/export_container.cpp
+++ b/BambooTracker/chips/export_container.cpp
@@ -1,4 +1,5 @@
 #include "export_container.hpp"
+#include "export_handler.hpp"
 #include <algorithm>
 
 namespace chip
@@ -37,8 +38,9 @@ namespace chip
 	}
 
 	//******************************//
-	VgmExportContainer::VgmExportContainer(uint32_t intrRate)
-		: lastWait_(0),
+	VgmExportContainer::VgmExportContainer(int target, uint32_t intrRate)
+		: target_(target),
+		  lastWait_(0),
 		  totalSampCnt_(0),
 		  intrRate_(intrRate),
 		  isSetLoop_(false),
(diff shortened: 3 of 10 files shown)


jpcima:
This is correct. When it comes to register-level, I have not done the inventory of what bits may or may not be in use, and considered it fine (probably?) if they pass through regardless.
I have set up the filter for cases where it saves space in the file.

I believe also that not all have the flag "AM enable" for example.

jpcima:
Judging from a look at MAME code, this register will surely interfere, so yes I'll mask it.

rerrahkr:
It works well, thanks for your contribution!

OPNA2608:
In hindsight, I don't know how I feel about this. I don't think it's really within the scope of this project to support the YM2612 or standalone AY-3-8910, especially when there are far more capable trackers for those already.

cyberic99:
> especially when there are far more capable trackers for those

I do not know any working tracker for YM2612, do you have a link?

For YM2149, there are some native trackers on Atari ST, sur as Maxymizer, MusicMon, Checktrakk...

But there are not a lot of modern trackers working on the 3 major platforms...
I only know about Archos Tracker 2
And it is not open source

So... I too think that there is still room for BambooTracker

(I could have forgotten some other trackers I do not know about... feel free to complete my comment)

jpcima:
>support the YM2612 or standalone AYs, especially when there are far more capable trackers for those already?

Personally, I've wanted to research for a workflow for AY tunes which favors use of free software only.
With such an option, it's possible.

On a Garvuino with atmega AY emulator synth, it opens a possibility. Previously, only Arkos Tacker and Vortex Tracker II have been available for this work.
BambooTracker has received a mention in the [Garvuino manual](https://github.com/farvardin/garvuino/blob/4d7331d8405181ec60dda9ae2a3b9744584d01a7/doc/garvuino_manual.t2t#L234).

Also the compatibility within the OPN family is nice to take advantage of; Mucom88's example is to consider, which has been used for Megadrive music.

papiezak:
>I do not know any working tracker for YM2612, do you have a link?

@cyberic99 There is DefleMask. Available on Mac, Windows, Linux, and very recently iOS and Android. Closed-source, development seems to have stopped around 2016, except mobile versions. Supports VGM logging, CH3 special mode, has MIDI input feature. It also supports YM2151, SID( a bit glitchy), Gameboy sound hardware, Master System's PSG, etc...

cyberic99:
> There is DefleMask

Yes right.

I knew it could export to ym2151 but forgot about ym2612.

The devs seem to promote the mobile version, but as you say the desktop version is lagging.
And there are many bugs.

If deflemask was open source, I think that many f these bugs would have been fixed and that it could have received a lot of help from the community.

And Deflemask doesn't support ym2149.

Deflemask has a lot of potential, and changing the systems while keeping the notes data is very useful.

But as of today, is is no more maintained.
And I find the UI confusing sometimes, and some actions are a bit strange (like using the middle click button), or accidental click increases the column number...

OPNA2608:
> There is DefleMask

> development seems to have stopped around 2016, except mobile versions

> But as of today, is is no more maintained.

[DefleMask](http://deflemask.com/) is still in active development afaik. As we're speaking, they're doing a poll on what system to ~bloat their software package with~ **add next** :confused:. It's not open-source yet, but they aim for going there "eventually". In the future. Supposedly.

I don't know where you all got the idea that it has seized development?

> For YM2149
> But there are not a lot of modern trackers working on the 3 major platforms...

YM2149 and AY stuff in general? Anyone I know that composes for those kind of chips either use native trackers, Arkos, Vortex II or 0CC-FamiTracker with the S5B expansion. Those who have tried BambooTracker for this use-case found it less useful command-wise than the alternatives. Furthermore, VGM is not exactly the most popular AY song format.

> Also the compatibility within the OPN family is nice to take advantage of;

But does BambooTracker really *need* to become the next DefleMask, a "jack of all trades, master of none" wanting to support *all teh chips*? Is aiming for universal OPN-family support really within the scope of BambooTracker :confused:? @rerrahkr 

> BambooTracker is a music tracker for the Yamaha YM2608 (OPNA) sound chip which was used in NEC PC-8801/9801 series computers.

papiezak:
>DefleMask is still in active development afaik. As we're speaking, they're doing a poll on what system to bloat their software package with add next

The same poll was over three years ago, OPL2/OPL3 has won if I remember correctly 🤣 I don't think devs added anything like that, same with many other glorious features  🙄 

>But does BambooTracker really need to become the next DefleMask, a "jack of all trades, master of none" wanting to support all teh chips?

Well, IMO BambooTracker should perfect and polish YM2608 system support as much as it's possible, then _maaaaaybe_ add YM2203 or anything else.

pedipanol:
>But does BambooTracker really need to become the next DefleMask, a "jack of all trades, master of none" wanting to support all teh chips? Is aiming for universal OPN-family support really within the scope of BambooTracker 😕?

@OPNA2608 The reason I requested other soundchip exports was because YM2203 (note that I didn't request YM2612) and YM2419 export could be implemented without interfering on the focus of the tracker. I didn't mean for it to become jack of all trades, just for it to become a master of OPNA while giving other options that wouldn't disrupt the workflow of the tracker itself.

It would be simply exporting the VGM files in other compatible formats, no change in the core of the tracker itself. Which is different from deflemask's multichip support, that requires a complete change in emulation for each extra soundchip.

This is why I dismissed when someone mentioned OPM support, it's a completely different soundchip that would require a lot of work to implement.

I _did_ request the ability to hide unused channels, but it's just that: hide. It's something I want even for making OPNA stuff, considering I personally barely use Rhythm.

jpcima:
@OPNA2608 it's as @pedipanol said above, exporting is very different, and I am also entirely against the idea of having the multi-chip. I think to characterize the feature as such is a vast exaggeration.
Possibly, it may belong in some "advanced" section out of the view of regular users; without too much doubt, it remains a marginal use case.

OPNA2608:
> (also YM2203 was a soundchip used in the NEC PC-x801 series)

Hence the YM2203 is the only one I think we should really target in addition to the YM2608. Along with volume presets to simulate the differences between the PC-9801-86 and SpeakBoard, it would cover all of the PC-9800 series' sound options most people would care about. (Beeper and generic Windows-era PCM audio aside).

But it shouldn't be a priority for now, as you can easily simulate the YM2203's feature set by just not using FM4-6, ignoring the Rhythm module and sticking to Mono panning, whereas the YM2608 in BambooTracker is still lacking ADPCM capabilities.

> I think to characterize the feature as such is a vast exaggeration.

It sends the mixed signal of BambooTracker now also supporting / working on supporting the YM2612, I saw discussions like that on Discord servers I use and on Twitter.

rerrahkr:
The reason I began creating BambooTracker is that there was no tracker of YM2608 (and YM2203) as far as I know, so I think YM2608 is the principal target of the tracker. Since YM2608 have YM2203 mode, it is not difficult to support YM2203.
I think YM2608 is an extension of YM2203, however, since the YM2608 and YM2612 have differences in SSG and PCM, different emulations are required. In addition, there is also a YM2612 tracker, so I will not do full support for YM2612. However it were able to export FM and SSG channels to YM2612 vgm at the moment, I added only limited export function.

jpcima:
>It sends the mixed signal of BambooTracker now also supporting / working on supporting the YM2612, I saw discussions like that on Discord servers I use and on Twitter.

It's understandable why this can be a problem. I wouldn't mind really if YM2612 is hidden at present time.

1. about export limitations, I did my best effort to make this clear on the user interface, in eye catching manner. But the meaning is not explicited in textual way, a reason why users may interpret it with confusion.
2. on the YM2612 export specifically, the result will be only as intented when clock is matching to what BambooTracker uses; it would work on a software VGM player which can set clock as desired, but a user may expect it to work on Megadrive and it will fail, creating more confusion.
(There existed machines which run OPN2 at matching clock, but MD is not among them)

OPNA2608:
You can already convert YM2608 to YM2612 VGMs (FM-only ofc) and YM2149 (integrated) to AY, the tools required are public and work on Windows and Linux as well, all it took on Linux was to disable the stack smashing detection for the main executable to run. One of the tools is an f-number converter to a clock rate of your choice to make them hardware-compatible too.
https://youtu.be/Yr5-2-w5eRE

jpcima:
@OPNA2608 Well yes, I'm well aware, but the approach is imperfect and has timbre distorsion.

PR #103, opened 2019-04-14
Visual

jpcima:
I try to add a nice visual of the generated waveform.
 
I'm storing the samples onto a circular buffer up to the capacity 1024. (~18.5 ms of signal)
When it's not busy, it will be transfered on a secondary buffer for thread-safe use.

The second commit adds the graphical widget.
The display is functional and nice, but I don't know too much how to lay it out without eating too much space. In this regard, current form is not very pleasing and should be rearranged.

### DIFF
--- a/BambooTracker/gui/wave_visual.cpp
+++ b/BambooTracker/gui/wave_visual.cpp
@@ -1,0 +1,76 @@
+#include "gui/wave_visual.hpp"
+#include "gui/color_palette.hpp"
+#include <limits>
+#include <QPainter>
+#include <QDebug>
+
+WaveVisual::WaveVisual(QWidget *parent)
+	: QWidget(parent)
+{
+	setAttribute(Qt::WA_OpaquePaintEvent);
+}
+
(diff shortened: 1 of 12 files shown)


BambooTracker/gui/wave_visual.cpp
```diff
@@ -0,0 +1,75 @@
...
+	const int16_t *samples = samples_.data();
+	size_t frames = samples_.size();
+
+	painter.setPen(palette->wavDrawColor);
+
+	int lastY = h / 2;
+	for (int x = 0; x < w; ++x) {
+		size_t index = (size_t)(x * ((double)frames / w));
+
+		int16_t sample = samples[index];
```
rerrahkr: crash when initial WaveVisual update is happened and samples isn't set to `samples_`. It needs buffer size check before plotting.

jpcima: Added the check `if (frames <= 0)` which prevents entering in this case.

rerrahkr:
Thanks for your contribution!

PR #105, opened 2019-04-14
send signal back into output history, not front

jpcima:
I've just realized it's the back samples which should be sent in the history buffer, in case the buffer size used is larger than history.
If there is some variance in the requested block size, then a discontinuity could be seen.


rerrahkr:
Thanks for your contribution!

PR #106, opened 2019-04-15
Chip updates

jpcima:
Add some minor things of little consequence.

- first is a backport of https://github.com/mamedev/mame/commit/f7465887710430dca424817bd54dc6df71c5d02c, a recent fix I have stumbled upon as I checked out updated sources
- second is a dumb logic error in Nuked-Mod I made, which prevents the PSG reading code to be ever entered, due to the value of `chip_type`. this is not of importance in context of BT, but might serve someone interested to reuse this code.


rerrahkr:
Thanks for your contribution!

Issue #32, opened 2018-12-16
MIDI keyboard support


galap-1:
A good feature to add would be support for using MIDI keyboards to enter notes.

rerrahkr:
Thank you for the request.
I will work on it.

jpcima:
Other ideas about MIDI input:
- synchronizing tempo to external clock
- real-time messages / MMC: start, stop, seek

Do you mind that I begin work to implement MIDI In? (for note inputs)
I'm very interested in having this.

rerrahkr:
Thanks @jpcima, it's very helpful!

jpcima:
I can have BambooTracker receive MIDI messages in the [work branch](https://github.com/jpcima/BambooTracker/tree/midi).

Implemented:
- callback MIDI input
- physical port, virtual port (only non-Windows OS)
- configuration dialog "MIDI" tab
- saving and loading connected device from settings

Then it's needed to make it react on input events.
I thought the simplest is to transmit desired events to Qt's event handling, and handle similarly as tone input from QKeyEvents.

>I had a look around some time ago for promising cross-platform C(++) libraries for MIDI I/O

I picked rtmidi, I'm long time user and contributor of this library.
None of portability libraries are perfect honestly, this should do fine for the task.

jpcima:
Notes about MIDI support in branch #76 

- has RtMidi in `midi/RtMidi/` subfolder. The origin is [this repo](https://github.com/Wohlstand/RtMidi), which has know fixes, notably ALSA, which upstream is slow to adopt.
- in `midi/` is a single instance helper class. It registers a callback set to receive events, the call arrives from a dedicated thread.
- the callback has two registrations; in `MainWindow` for jam mode, `PatternEditorPanel` for editing.
- the event is transmitted to the UI thread; it pushes a slot invocation into the event loop by `QueuedConnection`.

I have not enabled the JACK support in RtMidi.
JACK is hard-realtime sensitive, but this fashion of MIDI processing is non-RT.

galap-1:
Ah, it's really great to see this implemented!

PR #107, opened 2019-04-15
MIDI: selection of instrument by program change message

jpcima:
#32
This handles MIDI program change.
The message selects the instrument with the same number from the instrument list.


rerrahkr:
Thanks for your contribution!

Issue #96, opened 2019-04-10
MacOS Support


OPNA2608:
I was approached by a macOS user on Discord recently about getting BambooTracker to run natively on macOS. While it supposedly does run well in Wine, afew things still cause crashes and a native .dmg image for installation is just more appealing than fiddling with Wine.

I think it therefore makes sense to invest some time into getting it running on macOS and providing a precompiled binary via CI, as manual compilation is an unreasonable expectation for alot of end-users of this platform (downloading Xcode, the massive Qt5 framework, working in the Terminal, …) and, as of right now, shouldn't even succeed.

---

#### Points of Interest / Problems:
- ~macOS (or Xcode?) comes with the clang compiler instead of gcc, we should check if the argument and strictness differences between these two is significant enough to abort the compilation prematurely
(I tried compiling BambooTracker with clang on Linux some time ago and afaik it only failed on the very last compilation step with `unknown option: -stdlib=libstdc++`, might be a system or `qmake` problem though)~
clang appears to work just fine, it just issues afew more warnings about unused varibales
- ~macOS does not support ALSA, instead it offers the CoreAudio framework for audio and MIDI as far as I understand~
this appears to be handled in the RtMidi code, see next comment for problems
- we'll need to setup Travis CI to build the code on macOS as well
- and the READMEs need to be expanded with a macOS section
- ~~and, of course, someone'd need to get a test system to develop and test all of these changes on. all of the Apple systems i own are way too old to get a modern-enough macOS version for Qt5 running (PowerMac G5, Late 2006 XServe), but i can see if i can eventually get a spare, more modern macOS system at work~~
I can get my hands on a macOS 10.11.3 machine at irregular intervals

OPNA2608:
I got a system running macOS 10.11.3, installed Xcode 7.3.1 and Qt 5.9.7 and attempted a compilation. Surprisingly, it got pretty far but eventually did end up failing.

I think something's not right with RtMidi on macOS. Maybe I'm missing something @jpcima? Take a look:

[BambooTracker.log](https://github.com/rerrahkr/BambooTracker/files/3063200/BambooTracker.log)
[...]

PR #108, opened 2019-04-15
Enhanced macOS support

OPNA2608:
See: https://github.com/rerrahkr/BambooTracker/issues/96#issuecomment-481628594
RtMidi, besides CoreMIDI, also requires the CoreAudio and
CoreFoundation frameworks to compile on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,25 @@
 language: cpp
-os: linux
-dist: xenial
+
+matrix:
+  include:
+    - os: linux
+      dist: xenial
+    - os: osx
+      osx_image: xcode8.3
+
+addons:
+  apt:
+    sources:
+      - sourceline: 'ppa:ubuntu-sdk-team/ppa'
+    packages:
+      - qt5-default
+      - qttools5-dev-tools
+      - qtmultimedia5-dev
+      - libqt5multimedia5-plugins
+      - libasound2-dev
+  homebrew:
+    packages:
+      - qt5
 
 git:
   depth: 3
@@ -10,10 +29,8 @@ branches:
     - master
 
 before_install:
-  - sudo add-apt-repository --yes ppa:ubuntu-sdk-team/ppa
-  - sudo apt update -qq
-  - sudo apt install qt5-default qttools5-dev-tools qtmultimedia5-dev libqt5multimedia5-plugins
-  - sudo apt install libasound2-dev
+  # osx: force linking of qt5 into PATH
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew link --force qt5; fi
 
 script:
   - cd BambooTracker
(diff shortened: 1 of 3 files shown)


OPNA2608:
I also have some changes for .travis.yml to test compilation on macOS I can commit in abit, though I'm unsure what version of Xcode should be targeted, as this also dictates the minimum supported macOS version of the final .app package.

rerrahkr:
I don't have mac, so I also don't know which version most people use. It is necessary to ask someone else.

OPNA2608:
The lowest Xcode version Travis CI offers is Xcode 6.4 for OS X 10.9 (October 22, 2013) and above (I think), but support for that has officially ended this January.

I tried Xcode 7.3 for OS X 10.11, but it complained about Travis CI's `brew` version being outdated.

Xcode 8.3 for macOS 10.12 (September 20th, 2016) is the oldest Xcode version I can compile BambooTracker with *on Travis CI*, so I'll commit a change to .travis.yml to add it.

---

That being said, Xcode 7.3.1 with Qt 5.9.7 on OS X 10.11.3 is the oldest combination I could get my hands on and it compiled fine, hence we should mention in the READMEs that it might be able to be built on even older OS X versions, but we can't offer a prebuilt image for those, only rough compilation instructions at best.

---

> I don't have mac, so I also don't know which version most people use. It is necessary to ask someone else.

Unfortunately, the used Xcode version also dictates some extra features the compiled app will always support like [dark mode with the latest version.](https://doc-snapshots.qt.io/qt5-5.12/macos.html#supported-versions)

OPNA2608:
Whoops, that's what I get for never rebasing my test branches :laughing: 

>gui/wave_visual.cpp:23:10: error: no member named 'abs' in namespace 'std'; did you mean simply 'abs'?
>                sum += std::abs(buffer[2 * i]);
>                       ^~~~~~~~
>                       abs
>/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/stdlib.h:129:6: note: 'abs' declared here
>int      abs(int) __pure2;
>         ^
>gui/wave_visual.cpp:24:10: error: no member named 'abs' in namespace 'std'; did you mean simply 'abs'?
>                sum -= std::abs(buffer[2 * i + 1]);
>                       ^~~~~~~~
>                       abs
>/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/stdlib.h:129:6: note: 'abs' declared here
>int      abs(int) __pure2;
>         ^
>2 errors generated.
>make: *** [wave_visual.o] Error 1
>The command "make" exited with 2.

[That appears to be an issue with libc++ not being fully C++11 compliant in that regard.](https://stackoverflow.com/a/30084734)
I'll try fixing that with a follow-up commit.

jpcima:
It's just a problem of missing include in this particular setting.
It's my fault for not including it, not a fault of libc++.
It's actually Mac is more rigorous on this, the comment is not very relevant.

rerrahkr:
Thanks for contributions!

PR #109, opened 2019-04-15
Set an icon on all the windows

jpcima:
I've noticed BT has the generic icon on every window.
This fixes the problem.


rerrahkr:
Thanks for your contribution!

PR #110, opened 2019-04-16
Re-layout the oscilloscope

jpcima:
I arranged the oscilloscope positioning, according to multiple attempts and personal preference.
The inconvenience is the reduction of height of the instruments panel.
Despite it, the program will be able to resize more easily to a smaller size.

Also, I add a setting which can turn on/off the display, like deflemask.
Some users may prefer to avoid a source of distraction.

### DIFF
--- a/BambooTracker/configuration.cpp
+++ b/BambooTracker/configuration.cpp
@@ -29,6 +29,7 @@ Configuration::Configuration()
 	retrieveChannelState_ = false;
 	enableTranslation_ = true;
 	showFMDetuneSigned_ = false;
+	showWaveVisual_ = true;
 
 	// Edit settings
 	pageJumpLength_ = 4;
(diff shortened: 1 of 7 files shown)


rerrahkr:
Is it possible to reduce the size of the oscilloscope a little more? It is favor to avoid making the instrument list too small, even when viewing an oscilloscope.

I'm trying to move Tick freq to the module settings dialog and delete the module settings dialog button (there is the same button in the toolbar). Furthermore song number and song title will be changed to 1 combobox like famitracker. I was thinking of putting the oscilloscope in a space left by these tasks..

anyway could you wait for a while until the cleanup is over?

jpcima:
Please adapt as you see preferrable. I will be patient. :wink:
For the size, I took initially deflemask as a reference, but dimensioning to any size will work fine.
Also the toggle edit 96acff2 can surely be cherry picked independently.

rerrahkr:
I changed module settings groupbox at 4f5bdb535a1caa27f3be29f390961a9c459339cf. I decided to leave the song title and number because it needs to change the song properties update order if they are changed to 1 combobox.

rerrahkr:
Since the function to hide the oscilloscope is convenient, I will merge this PR before a bit re-layout.

jpcima:
I remove from the PR the UI edit, and I leave the toggle setting.

rerrahkr:
Thanks for your contribution!

PR #111, opened 2019-04-16
Add export targets to S98

jpcima:
This is to complete the second part of task #98.

Now, S98 can be exported in the same way as VGM.
The file header can contain 2 device info blocks when a PSG is used.

The FM will map to command 00-01.
The PSG, if used, will map to 02 and 03(unused).


rerrahkr:
Thanks for your contribution!

Issue #113, opened 2019-04-19
Interpolation enters junk data


jpcima:
On ef71afc, sometimes interpolating will insert bad values into the pattern.
This seems only happening if the interpolated range is long enough, a shorter one is fine.

**Example**
![Capture du 2019-04-19 06-28-16](https://user-images.githubusercontent.com/17614485/56405394-5a26e800-626c-11e9-9241-4f84a546bd1d.png)

**Steps**
1. start with the empty song
2. at the volume column at row 00, enter `60`
3. at the volume column at row 30, enter `7F`
4. select the volume column 00 to 30 using shift key and arrows
5. press `ctrl+g` for interpolation

PR #114, opened 2019-04-19
Fix computed interpolation values

jpcima:
#113

Prevents the arithmetic promotion to unsigned. (because of `size_t`)


rerrahkr:
I missed the type check, thanks for your contribution!

PR #116, opened 2019-04-20
Position the cursor after setting line edit contents

jpcima:
This realigns text display to the left after setting the text edit.
It's in the case the box of the widget is too small for the contents.


rerrahkr:
Thanks for your contribution!

PR #117, opened 2019-04-20
Resolve warning 'catching polymorphic type by value'

jpcima:
It's a warning of GCC 8 which identifies the `catch` part as potential source of problems.


rerrahkr:
Thanks for your contribution!

Issue #49, opened 2019-01-02
Stuttery drawing of Pattern Editor


OPNA2608:
I'm experiencing a noticable stuttering of the Pattern Editor when playing a song in Follow Mode with any Buffer length >3ms.

[Here is a .WEBM showcasing this problem.](https://cdn.discordapp.com/attachments/410831172181164044/529843973343150100/PatternEditorStutter.webm)

(This is on Linux, NixOS 18.09, running the freshly compiled release v0.1.4)

rerrahkr:
Thanks for your report.
Due to the simultaneous updating of sound buffers and the pattern editor, it may occur.
I think that it is possible to avoid it by dividing these processes.

alexmyczko:
I can see the same here (debian or ubuntu)

rerrahkr:
It doens not happen on my Windows 10.
Is the sound correctly playing even at 3ms?

OPNA2608:
[Actual playback](https://github.com/rerrahkr/BambooTracker/files/2730516/recordings.tar.gz) is fine from what I can hear, with only minimal differences between 1ms and 500ms.

OPNA2608:
I think this issue has improved alot with the merging of #152. Though still noticeable at higher buffer lengths, I consider it closed for now (yet possibly still improvable!) as the main symptom (<3ms being required for usage) has been fixed. :slightly_smiling_face:

PR #120, opened 2019-04-27
Change audio output stream

rerrahkr:
#49, #96
I'm trying to change audio output stream.
Conventionally sample generation has been performed according to the remaining amount of buffer, however in this, sample generation is performed using a timer. The timer is set its interval to the module tick frequency. When the timer is raised, the tracker counts up tick, updates ui, generates samples for one tick and flushes them to a buffer.  And buffer filed, it send to `QIODevice`.
It seems to solve stuttering of ui drawings in this, but the sound is choppy. 😕


jpcima:
I have noticed about `QAudioOutput` as I tried to improve the audio situation:
it never uses `setBufferSize` on the audio object. Maybe increasing this buffer could help.

I believe the audio has essentially 2 problems:
- no tasks are separated, all are in a single thread
- the tick is on the Qt event loop and needs it to react fast, but there is not a notion of preemption or priority. so, ticks can be delayed by indefinite durations.

rerrahkr:
I close this draft to separate audio task from main graphic thread before fixing stuttering of drawings.
I'm going to use a library such as RtAudio to handle audio tasks in a separate thread.

jpcima:
I did work here a long ago, as an experiment, which ran audio in a separate thread.
It was unfinished, done as experiment, device selection was not done.

From what I remember of this experiment, this alone was not sufficient, it needed some deeper changes in the tracker engine itself, which I don't recall exactly from memory.

https://github.com/jpcima/BambooTracker/tree/audio-thread

rerrahkr:
Thanks for your information! I will check your branch.

Issue #122, opened 2019-05-06
Missing translations


trebmuh:
I've build a 2.0 version and am checking right now the French translation. Overall, it looks very good so far. Here below are a few missing translations.

1. Fichier (File) -> Ouvrir (Open) ->"Bambootracker module file" (file type selector) is not translated.
2. Same for Fichier (File) ->Sauvegarder sous (Save as)
3. Module -> Propriétés du module (module properties) : missing space for the "Numéro" column:
![bt-l10n-fr-1](https://user-images.githubusercontent.com/8705846/57261706-3b489400-7069-11e9-8a6b-c33395d42bb9.png)
4. Instrument -> Charger depuis un fichier... (Load from a file...) the 6 options from the selector aren't translated (Bambootracker instrument, Deflemask preset,...)
5. Instrument -> Importer depuis un fichier de banque... (Import from a bankfile...), the options from the selector isn't translated (WOPN bank)
6. The 3 submenu items' windows in the Aide (Help) menu aren't translated

_Pinging @jpcima_

jpcima:
I have submitted a PR for 1, 4, and 5. 
The items 2 and 6 are already translated, and should be part of the release 0.2.0, so it's surprising.

This leaves to examine the 3rd item.

Please note: I'm also aware of another text overflow in the exporting dialog, in case a wide font is used.
![foo](https://user-images.githubusercontent.com/17614485/57284412-68209980-70b1-11e9-805f-a338e183f3c6.png)

trebmuh:
>> 6. The 3 submenu items' windows in the Aide (Help) menu aren't translated
>  The items 2 and 6 are already translated, and should be part of the release 0.2.0, so it's surprising.

![liste_d_effets](https://user-images.githubusercontent.com/8705846/57293829-16363e80-70c6-11e9-847c-3e20d24cfd09.png)
![raccourcis_clavier](https://user-images.githubusercontent.com/8705846/57293833-1d5d4c80-70c6-11e9-9ae4-bb31686f65f2.png)
![a_propos](https://user-images.githubusercontent.com/8705846/57293837-1fbfa680-70c6-11e9-9b33-62479cce48a6.png)

jpcima:
The first 2 are moderately recent additions, and they're just not made. I did not took time to think of adequate vocabulary, etc.

[...]

PR #124, opened 2019-05-07
Automatic section size for table column

jpcima:
#122
Set a dynamic size on the table header.
It should allow the text to fit regardless of language and font size.


rerrahkr:
Thanks for your contribution!

Issue #130, opened 2019-05-31
Trouble with song export


AoiMimi:
Trying to export in any of the available formats (WAV, VGM, S98) makes the exported file immediately delete itself.
I tested this with the demo songs as well, and ran into the same problem.

papiezak:
That's weird, I can't replicate this problem, at least on Windows 10. Tried all demo songs, even my own one, every format, no success.
What OS are you using? Did you compiled tracker from source, or downloaded dev build from AppVeyor/0.2 release?

AoiMimi:
I'm using Windows 10 Home, and I've downloaded version 0.2.0 from the Releases tab.
Sorry I can't figure out what might be causing this either...

rerrahkr:
Thanks for your report. I could not replicated on my Windows 10. I should not have written the code to delete the file...

papiezak:
It can be a problem on the OS side, try running BambooTracker as an admin, exporting to several other directories or even flash drives, etc..
What about saving instrument or module files? They aren't being deleted after save?

AoiMimi:
Saving and loading instrument files works fine.
I tried running the program and exporting as admin, but still had the same issue, even with exporting to different directories.

papiezak:
Last thing you can do is trying on another PC or HDD/SSD, if you have access to it...

AoiMimi:
Until another solution is found, I will try that. Thank you!

scarletbullgon:
I have this same exact issue myself, but not only it occurs when the program isn't running under administration privileges, but also when the folder you're trying to export into has special characters with diacritics in them (for example; acutes such as á, ó, and diaeresis like ü).

jpcima:
Oh there is an encoding anomaly for Windows. It's dubious why it would remove files though.
BT passes utf-8 strings to file opening functions but this doesn't work, it takes a conversion to Utf-16 and calling the widechar variant. It's only Windows which needs having this workaround.

jpcima:
Can you report whether the version here helps solve the problem?
https://ci.appveyor.com/project/rerrahkr/bambootracker/builds/25017454/job/iev9gb9kjupmgh6b/artifacts

papiezak:
>when the folder you're trying to export into has special characters with diacritics in them

Now I'm able to replicate it, and #133 seems to fix it; thanks @jpcima!

rerrahkr:
I merged #133. Thanks all for reports and helps!

PR #133, opened 2019-06-03
Unicode paths

jpcima:
Attempt at bug #130

Windows Unicode paths

Problem: on Windows, `std::ifstream` and `std::ofstream` do not support unicode paths, except on MSVC using the non-standard widechar extension. Unfortunately, it's stupidly difficult on GCC, where one has to realize a custom stream.

I added headers from a library [nowide-standalone](https://github.com/nephatrine/nowide-standalone) which implements the custom stream. (on non-Windows, it will still use STL streams)
It's one solution with its own inconvenient, but working in my tests.

Secondly, I eliminate use of "Local8Bit" transformations, such that all paths transit in Unicode.


rerrahkr:
Thanks for your contribution!

PR #141, opened 2019-06-23
Add bank import and icons to instrument context menu

jpcima:
This sets icons in the right click menu. (the same way as the Instruments from menu bar)
Also it inserts the missing item Import from Bank File.


rerrahkr:
Thanks for your contribution!

PR #142, opened 2019-06-23
Add search box for instrument selection dialog

jpcima:
I added a search box to find instruments more easily inside a bank.

In the commit following I have updated the fr translation with 2 new strings.


rerrahkr:
Thanks for your contribution!

PR #152, opened 2019-09-22
Separate audio streaming task from Qt event loop

rerrahkr:
#96
Audio streaming task is separated from Qt event loop by using RtAudio. Based on jpcima@3e822d5d528d41185fef0269f8c397a2e355513b  (https://github.com/rerrahkr/BambooTracker/pull/120#issuecomment-526828245).

need a check on macos and linux


OPNA2608:
Feedback from Linux (NixOS 19.03):

1) This PR adds a dependency on *all* of the audio backends for Linux at the same time (PulseAudio and JACK2 are now `#include`d by RtAudio). This is a bad idea, ideally it should detect which backends are available during build time and include them accordingly. Alternatively, the user could pass a define for their backends to add support for them.

2) Attempting to use the `55466Hz` Sample Rate setting now results in this error message. The other 2 options work fine.
grafik

3) During playback, the pattern editor sometimes freezes up. It:
- stops scrolling during playback
- becomes unresponsive to input (scrolling does nothing, clicking only changes the highlighted channel in the order list)
- only moves to where the cursor is when playback is stopped and
- stays like this until BT is restarted.

4) Attempting to start with a sample rate that doesn't throw the error from 2. segfaults:
```
Translation "qt_de_DE" from "/nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/translations"
Translation "bamboo_tracker_de_DE" from "/home/bt1cn/devel/BambooTracker/BambooTracker/BambooTracker/../share/BambooTracker/lang"
Using emulator: MAME YM2608
RtAudio pulse: _NOT_ running realtime scheduling
==26038== Thread 6:
==26038== Use of uninitialised value of size 8
==26038==    at 0x4D9BC54: QLabel::setText(QString const&) (in /nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/lib/libQt5Widgets.so.5.12.0)
==26038==    by 0x42CBCB: MainWindow::onNewTickSignaled(int) (mainwindow.cpp:2441)
==26038==    by 0x5B9DE5E: QMetaObject::activate(QObject*, int, int, void**) (in /nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/lib/libQt5Core.so.5.12.0)
==26038==    by 0x5A957D: AudioStream::streamInterrupted(int) (moc_audio_stream.cpp:131)
==26038==    by 0x470842: AudioStream::tickNotifierRun() (audio_stream.cpp:122)
==26038==    by 0x5FA1D7E: ??? (in /nix/store/hlnxw4k6931bachvg5sv0cyaissimswb-gcc-7.4.0-lib/lib/libstdc++.so.6.0.24)
==26038==    by 0x5EC5EF6: start_thread (in /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27/lib/libpthread-2.27.so)
==26038==    by 0x650A22E: clone (in /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27/lib/libc-2.27.so)
==26038== 
==26038== Invalid read of size 8
==26038==    at 0x4D9BC54: QLabel::setText(QString const&) (in /nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/lib/libQt5Widgets.so.5.12.0)
==26038==    by 0x42CBCB: MainWindow::onNewTickSignaled(int) (mainwindow.cpp:2441)
==26038==    by 0x5B9DE5E: QMetaObject::activate(QObject*, int, int, void**) (in /nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/lib/libQt5Core.so.5.12.0)
==26038==    by 0x5A957D: AudioStream::streamInterrupted(int) (moc_audio_stream.cpp:131)
==26038==    by 0x470842: AudioStream::tickNotifierRun() (audio_stream.cpp:122)
==26038==    by 0x5FA1D7E: ??? (in /nix/store/hlnxw4k6931bachvg5sv0cyaissimswb-gcc-7.4.0-lib/lib/libstdc++.so.6.0.24)
==26038==    by 0x5EC5EF6: start_thread (in /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27/lib/libpthread-2.27.so)
==26038==    by 0x650A22E: clone (in /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27/lib/libc-2.27.so)
==26038==  Address 0x8 is not stack'd, malloc'd or (recently) free'd
==26038== 
==26038== 
==26038== Process terminating with default action of signal 11 (SIGSEGV)
==26038==  Access not within mapped region at address 0x8
==26038==    at 0x4D9BC54: QLabel::setText(QString const&) (in /nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/lib/libQt5Widgets.so.5.12.0)
==26038==    by 0x42CBCB: MainWindow::onNewTickSignaled(int) (mainwindow.cpp:2441)
==26038==    by 0x5B9DE5E: QMetaObject::activate(QObject*, int, int, void**) (in /nix/store/c7cmk9ikb3gl3arysl7sm9livpml1zpc-qtbase-5.12.0/lib/libQt5Core.so.5.12.0)
==26038==    by 0x5A957D: AudioStream::streamInterrupted(int) (moc_audio_stream.cpp:131)
==26038==    by 0x470842: AudioStream::tickNotifierRun() (audio_stream.cpp:122)
==26038==    by 0x5FA1D7E: ??? (in /nix/store/hlnxw4k6931bachvg5sv0cyaissimswb-gcc-7.4.0-lib/lib/libstdc++.so.6.0.24)
==26038==    by 0x5EC5EF6: start_thread (in /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27/lib/libpthread-2.27.so)
==26038==    by 0x650A22E: clone (in /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27/lib/libc-2.27.so)
```

5. The README sections on manual compilation should be updated to reflect the changed dependencies.

---

I'll see when I can test it on a macOS system. Thanks to jpcima and you for working on this problem. :slightly_smiling_face: :tada:

rerrahkr:
Thank you @OPNA2608 for checking!

> 1\. This PR adds a dependency on _all_ of the audio backends for Linux at the same time

To be honest, I just used jpcima `RtAudio.pri` and I don't know what these backends are different.

```qmake
linux {
    DEFINES += __LINUX_ALSA__
    LIBS += -lasound -lpthread
    DEFINES += __LINUX_PULSE__
    LIBS += -lpulse-simple -lpulse
    DEFINES += __UNIX_JACK__
    LIBS += -ljack
}
```

Does linux make audio even if only one of these is used?

> 2\. Attempting to use the `55466Hz` Sample Rate setting now results in this error message. The other 2 options work fine.

It is the error that RtAudio could not open the stream the stream with the specified configuration, sample rate is 55466Hz. I've noticed this error when building on macos before. Since this error could not be avoided, I have the tracker show this dialog to let the user change the settings, such as dropping the sample rate.


> 3\. During playback, the pattern editor sometimes freezes up.

I find that it is occurred when speed is 1, it seems something is wrong about mutex lock of playing position.

> 4\. Attempting to start with a sample rate that doesn't throw the error from 2.

It is the initialization error, I fixed it at 6bddf50.

OPNA2608:
Feedback from macOS (OS X El Capitan, 10.11.6):

1. Attempting to change the window to fullscreen makes the window unresponsive to any input and stay windowed.
If playback ceases on its own though (by reaching the end of the track without looping), it then switches to fullscreen and becomes full responsive again. macOS does not report the window as "not responding" during that time.

2. If the end of a song has been played, either during live playback or recording, and the user attempts to restart playback / load another module or the recording part tries to finish, the app becomes unresponsive. (macOS beach ball thingy shows)

(I don't know if those problems are exactly *caused* by this PR, but they're things I noticed while testing it.)

The general lagginess of the UI during playback remains, but the sound itself is fine now. :+1:

---

> I don't know what these backends are different.
> Does linux make audio even if only one of these is used?

Put briefly, ALSA is the lowest-level sound backend.
Both PulseAudio and JACK are working on-top of ALSA for more sophisticated/user-friendly sound management, and they have their pros and cons each. You usually run only either of them at a time, or use a shim to pass one of them through the other.

You're able to use ALSA on its own, PulseAudio (+ ALSA) or JACK (+ ALSA) to get sound on Linux.

(I don't know how RtAudio decides which one to use, but maybe having a SelectBox in the configuration dialogue to pick which of the currently available backends to use would be useful)

If testing the backends on their own is as easy as commenting out the appropriate `DEFINES` and `LIBS` from that pri, I can try doing that later and report back.


> Since this error could not be avoided, I have the tracker show this dialog to let the user change the settings, such as dropping the sample rate.

Exporting to WAV with a sample rate of 55466Hz appears to work fine under macOS (file reports that sample rate, anyway), maybe you could change the dialogue to a warning that with the selected sample rate, only exporting is possible.

rerrahkr:
Thanks for testing on macos! I will check your report when I can use macos.

> If testing the backends on their own is as easy as commenting out the appropriate `DEFINES` and `LIBS` from that pri, I can try doing that later and report back.

Yes, you could select the backends by commenting out `DEFINES` and `LIBS`. Could you test them?

About WAV export, I have just added the sample rate selection in the WAV export dialog at 64e4c02baa938628486299d6580d2cb8feef5cb5, so the sample rate of wav does not depend on the audio stream.

OPNA2608:
> Yes, you could select the backends by commenting out DEFINES and LIBS. Could you test them?

I can confirm ALSA-only and PulseAudio working by soloing the `__LINUX_ALSA__` and `__LINUX_PULSE__` `DEFINES` and `LIBS`. I was able to compile the soloed JACK version, but I don't have JACK2 setup on my system so it fell back to ALSA.

> 2. If the end of a song has been played, either during live playback or recording, and the user attempts to restart playback / load another module or the recording part tries to finish, the app becomes unresponsive. (macOS beach ball thingy shows)

This actually also appears to happen on Linux, as I've just noticed.

rerrahkr:
> 2\. If the end of a song has been played, either during live playback or recording, and the user attempts to restart playback / load another module or the recording part tries to finish, the app becomes unresponsive. (macOS beach ball thingy shows)

I fixed it at 37296f3, the bug was the undesirable lock of tick counter.

I think to change `RtAudio.pri` about `DEFINES` on linux:
```qmake
linux {
   defined(__LINUX_PULSE__) {
       # DEFINES += __LINUX_PULSE__
       LIBS += -lpthread -lpulse-simple -lpulse
   }
   else:defined(__UNIX_JACK__) {
       # DEFINES += __UNIX_JACK__
       LIBS += -ljack -lpthread
   }
   else {
       DEFINES += __LINUX_ALSA__
       LIBS += -lasound -lpthread
   }
}
```
Backends can be selected by adding an option when executing qmake. For example, if you would like to use PulseAudio, you run qmake as `qmake DEFINES+=__LINUX_PULSE__`.

rerrahkr:
I checked the Qt event “emit” and found that frequent calls to the pattern editor drawing events with high processing costs caused these problems. It is necessary to reduce the count of event call and drawing costs.

BambooTracker/stream/RtAudio/RtAudio.pri
```diff
@@ -5,24 +5,37 @@ HEADERS += \
...
+        LIBS += -lpthread -lpulse-simple -lpulse
+    }
+    else:defined(__UNIX_JACK__) {
+        # DEFINES += __UNIX_JACK__
+        LIBS += -ljack -lpthread
+    }
+    else {
+        DEFINES += __LINUX_ALSA__
+        LIBS += -lasound -lpthread
+    }
```
OPNA2608: Sorry for not testing this earlier, but this actually doesn't work. Running
```bash
qmake DEFINES+=__LINUX_PULSE__
```
adds `__LINUX_PULSE__` to `DEFINES`, but falls into the else statement here.

Looking at the [qmake documentation](https://doc.qt.io/qt-5/qmake-test-function-reference.html#defined-name-type), `defined` tests if a *variable called \<x\>* exists, not if a flag has been added to `DEFINES`.

I suggest [`contains`](https://doc.qt.io/qt-5/qmake-test-function-reference.html#contains-variablename-value) instead:
```qmake
contains(DEFINES,__LINUX_PULSE__) {
```
etc.

jpcima:
Hi. It's fine enabling RtAudio multiple backends at the same time.
At runtime, RtAudio is able to pick one which works the best on a given system.

I usually enable all three, it makes a configuration which is compatible with about all systems.
- PulseAudio is what most users have, which is essentially a software mixer like Windows or Mac's
- ALSA is the most low-level form of audio
- JACK is for pro-audio users which are interested in plugging software together, for instance to give BambooTracker an effect chain. It is useful for enabling such possibilities.

rerrahkr:
Is it better to enable to select multiple `DEFINES` than only 1 backends, like:
```qmake
linux {
    contains(DEFINES, __LINUX_PULSE__) {
        # DEFINES += __LINUX_PULSE__
        LIBS += -lpthread -lpulse-simple -lpulse
    }
    contains(DEFINES, __UNIX_JACK__) {
        # DEFINES += __UNIX_JACK__
        LIBS += -ljack -lpthread
    }
    DEFINES += __LINUX_ALSA__
    LIBS += -lasound -lpthread
}
```

Since I find `RtAudio::getCompiledApi(std::vector<RtAudio::Api> &apis)` and the constructor with the argument of specific `RtAudio::Api`, I will add backend selection in configuration dialog.

rerrahkr:
The purpose of this PR is to separate audio tasks from Qt event loop, so if there is no problem, I will be merged.
The delay of GUI drawing was confirmed at `master` on macos. I will work on it by creating a new branch.

OPNA2608:
LGTM. :tada:

rerrahkr:
Thank you for your help!

PR #155, opened 2019-10-06
Reduce drawing cost of pattern editor and order list

rerrahkr:
#96, #152
- Separate the pixmap of these editors each layer of pixmap (row foreground, background, header) and combine them when drawing. This eliminates unnecessary repainting. For example, when pattern data is changed, only the foreground of rows needs to be redrawn.
- Change order list update timing from step down to order change.
- Shift and reuse existing rows in step down and order row down. The area that needs to be painted becomes smaller.
- Disable mouse slider operation related to changing the cursor during playback.
- Replace drawing call from `update()` to `repaint()`. `repaint()` performs an immediate repaint.

On Windows 10 and macos Mojave (10.14.6), the pattern editor, order list and wave display seems not to stuck.


OPNA2608:
Wonderful, thanks! :+1:

> On Windows 10 and macos Mojave (10.14.6), the pattern editor, order list and wave display seems not to stuck.

On OS X El Capitan (10.11.6) here, the wave display starts getting laggy when moving around the mouse over the pattern editor during playback but is fine otherwise.

There seems to be (a few?) ways to achieve a recursive repaint with this. Attempting to open another module while a module is currently playing froze the window completely sometimes, or brought up an endless amount of windows with this error message other times.

Bildschirmfoto 2019-10-07 um 13 55 10

One time, this seemingly leaked into the BambooTracker.ini and instantly produced this error message on startup, requiring its removal to be fixed. (Unfortunately, I drew the connection too late and deleted it instead of moving it somewhere safe for examination :slightly_frowning_face:)

In either case, after a while, the following was printed in the Terminal:

> QWidget::repaint: Recursive repaint detected

I can't reproduce it every time, though it seems to happen more frequently when playback is about to jump into the following pattern. Hope this helps!

OPNA2608:
Unsure if this is helpful, but here's an [lldb backtrace](https://github.com/rerrahkr/BambooTracker/files/3697280/backtrace.txt) while the application was frozen.

OPNA2608:
Same error (I think) on Linux, this time with a full error message:

grafik

Appears to be much simpler than expected, if it's the same bug. To reproduce (on Linux at least):
1. open `breeze_2608.btm`
2. go to any pattern with a size >= `0x40`
3. move the cursor to an offset >= `0x40` and
4. attempt to open `Neo Megalopolis.btm`.

Can't reproduce on `master`.

rerrahkr:
Thank you for checking!

> the wave display starts getting laggy when moving around the mouse over the pattern editor during playback, as do the pattern editor and order list when scrolling horizontally

It is the reason that these operation needs to repaint background many times.

Recursive repaint and vector size error could be caused by changing `update()` to `repaint()`.
`update()` is executed in the main event loop of Qt, but `repaint()` is independent, so it seems that `repaint()` is called before module loading has been finished and it happens an error.

Could you try to check 9bfe3f5 that repaint requests by `update()` on macos and linux? If there is no problem, I will use `update()`. At least on my windows 10 I didn't see any significant difference due to the change.

rerrahkr:
I changed the pattern editor at 1483533, to repaint only the background and reuse foreground text when moving the cursor to the column in displayed track.

rerrahkr:
At b516ebf, the repaint of the pattern editor and the order list that need to see the module data is skipped during loading module and changing song.

If there is no problem, I will merge this PR and change .travis.yml to deploy the binary for macos.

rerrahkr:
Fixed some mistakes at 3fcb635.

OPNA2608:
Apologies for not responding.
I won't have access to a macOS system again for a while, but it looks fine on Linux.

rerrahkr:
I tested on macos Mojave (10.14.6) and it looks fine too. Therefore I will merge this PR.

rerrahkr:
Thank you for your help!

Issue #172, opened 2019-12-16
Error creating ALSA sequencer client object.


OPNA2608:
On my main system (Linux, NixOS 19.09), after a fresh boot, I'm unable to start BambooTracker. It aborts with a SIGSEGV and prints this:
```
Translation "qt_de_DE" from "/nix/store/p16qsg4ram58g9aalkd0dixf000jgzbh-qtbase-5.12.3/translations"
Translation "bamboo_tracker_de_DE" from "/home/bt1cn/devel/BambooTracker/BambooTracker/BambooTracker/.qm"
Using emulator: Nuked OPN-Mod
ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: Permission denied

MidiInAlsa::initialize: error creating ALSA sequencer client object.

Segmentation fault (Speicherabzug geschrieben)
```
(last line means process aborted with a segmentation fault)

If I start the app with `sudo` once and try again, it works fine. Same if I add myself to the `audio` group and log out and back in.
It's possible my system is misconfigured and I would indeed need the `audio` group or proper permissions for /dev/snd/seq, however I have configured the use of the Pulse backend instead so the error is a bit irritation. I guess this comes from the MIDI initialization? Furthermore, on my Ubuntu machine I'm not in the `audio` group either yet it starts up just fine there.

Here's the debugger report on the actual SIGSEGV:
```
MidiInAlsa::initialize: error creating ALSA sequencer client object.

==6731== Invalid read of size 8
==6731==    at 0x5257E08: ??? (in /nix/store/p16qsg4ram58g9aalkd0dixf000jgzbh-qtbase-5.12.3/lib/libQt5Gui.so.5.12.3)
==6731==    by 0x5259204: QGuiApplication::font() (in /nix/store/p16qsg4ram58g9aalkd0dixf000jgzbh-qtbase-5.12.3/lib/libQt5Gui.so.5.12.3)
==6731==    by 0x5311770: QFont::QFont() (in /nix/store/p16qsg4ram58g9aalkd0dixf000jgzbh-qtbase-5.12.3/lib/libQt5Gui.so.5.12.3)
==6731==    by 0x4C5B580: QWidgetPrivate::QWidgetPrivate(int) (in /nix/store/p16qsg4ram58g9aalkd0dixf000jgzbh-qtbase-5.12.3/lib/libQt5Widgets.so.5.12.3)
[...]

PR #173, opened 2019-12-17
Fallback in case system MIDI cannot be opened

jpcima:
#172

This prevents BT crashing in case the MIDI system fails to open for any reason.

Verified as working on Linux ALSA.
I have set zero permissions on the device node for testing.

It needs to force rebuild RtMidi.cpp with a definition which enables dummy mode.


rerrahkr:

PR #183, opened 2020-02-23
Nuked ADPCM

jpcima:
Hi, this is ADPCM implemented into Nuked OPN.
Let me know if this works well.

The last buffering issue has been resolved by adding a method which is able flush the register buffer, and this operation will be performed prior to device resets.


rerrahkr:
It works well as far as I tested, I will merge this PR. Thanks for your great contribution!

PR #204, opened 2020-04-13
Fix missing install target for translations

OPNA2608:
See QTBUG-77398, https://bugreports.qt.io/browse/QTBUG-77398. Affects
any Qt5 version from 5.12 to 5.14.1, which some Linux distros still use.

Before, without any .qm files in `BambooTracker/.qm` while invoking `qmake`:
install_qms_before

After:
install_qms_after


rerrahkr:
Thanks for your contribution!

PR #205, opened 2020-04-13
README adjustments

### LINKED ISSUE #203: Missing qt dependency for Linux
Perhaps this issue only pertains to Debian, but I had to install the package `qttools5-dev-tools` because the `lrelease` binary was missing from the qt directory. I would suggest adding it, or perhaps mentioning it, since other Debian-based distros may face the same requirement.

OPNA2608:
Fixes #203.

---

I tried to expand on the whole Building section abit with more general information, with subsections for the different OS'/Distros. I can't say too much on the Windows & macOS sides of things, feel free to expand on those sections yourself & push to the branch or tell me what to put there. And the Japanese version of the README will need adjustments as well, I'd assume.

With the [Wiki](https://github.com/rerrahkr/BambooTracker/wiki) being available & less in-the-way, could we strip the READMEs of these sections and, if it hasn't been done yet, move all that information to the wiki btw? I think it'd increase the readability of the README quite abit.
- Glossary
- Interface Overview
- Key commands
- Volume range
- Effect list
- File I/O

Also, maybe some information in the README on how to contribute to BambooTracker would be good - i18n, skins, demo modules etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,11 @@ To build BambooTracker, you'll need the following required dependencies:
 
 The way of acquiring all these will depend on your OS & distribution.
 
+#### Windows
+**TODO**
+- MinGW / MSVC
+- Qt5
+
 #### macOS
 You'll most likely need to install the Xcode Command Line Tools - how to acquire these might depend on your macOS version.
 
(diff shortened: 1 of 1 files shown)


rerrahkr:
Thanks for your contribution! I'll also change some texts of READMEs.

PR #221, opened 2020-06-13
Further modify compile settings

OPNA2608:
As requested in #218, this merges https://github.com/OPNA2608/BambooTracker/commit/d534a18 with `master`'s changes.

Issue #218, opened 2020-05-24
Improve C++ standard compliance and reduce pedantic warnings


OPNA2608:
Related to #216.

BambooTracker has some snippets of code that a compiler set to be very pedantic about standard compliance and good practices will quickly point out and error on. I thought that a meta-issue about fixing them and how to ensure a compliant codebase in the future would be useful.

Some of those problems stem from included projects, I decided to point them out regardless. We could fix them only for ourselves / submit any changes upstream or ignore them for the unforeseeable future.

---

#### How to verify the problems (or: how to get your compiler to nag about compliance and bad practices):

On GCC/MinGW and Clang:

```sh
qmake (the usual options here) QMAKE_CXXFLAGS+="-Wall -Wextra -Werror -pedantic -pedantic-errors (+ whatever exceptions are required to compile, ideally none)"
```
Or set the `QMAKE_CXXFLAGS` setting in `BambooTracker/BambooTracker.pro` I had to set the same setting in `BambooTracker/stream/RtAudio/RtAudio.pri` and `BambooTracker/midi/RtMidi/RtMidi.pri` for it to work.
(If you know a better way of setting this, please let me know!)

MSVC appears to have no flag to enable pedantic checking, but it has some variations of `/W…` ones to enable extra warnings and mark warnings as errors.

#### Currently needed exceptions required to compile successfully after some easy fixes (tested on Linux, GCC & Clang):

(Prefix to disable a specific warning is `-Wno-error=`)

- `unused-parameter` (low priority)
`BambooTracker/chips/c86ctl/c86ctl_wrapper.cpp` - several methods
- `deprecated-declarations` (Clang-only, low prioriy)
`BambooTracker/midi/RtMidi/RtMidi.cpp, several methods - `'(method declaration)' is deprecated`
- `redundant-move` (Clang-only)
`BambooTracker/instrument/instrument.cpp`, several methods - `redundant move in return statement`
- `type-limits`
`BambooTracker/chips/codec/ymb_codec.hpp`, line 14 - `comparison of unsigned expression < 0 is always false`
- `maybe-uninitialized`
`BambooTracker/playback.cpp` line 336 - `»storeEffectToMap.bool (PlaybackManager::*)(int, Effect)::__pfn« may be uninitialised`
`BambooTracker/command/pattern/set_effect_value_to_step_command.cpp`, line 35 - `'value' may be used uninitialized in this function`
- `vla`
[...]

PR #227, opened 2020-06-23
Fix leftover compilation warnings

rerrahkr:
#218
Try to remove several compilation errors left from #221.

### DIFF
--- a/BambooTracker/midi/RtMidi/RtMidi.pri
+++ b/BambooTracker/midi/RtMidi/RtMidi.pri
@@ -5,6 +5,10 @@ HEADERS += \
     $$PWD/RtMidi.hpp
 
 linux {
+    contains(DEFINES, __UNIX_JACK__) {
+        DEFINES += JACK_HAS_PORT_RENAME
+        LIBS += -ljack
+    }
     DEFINES += __LINUX_ALSA__
     LIBS += -lasound
 }
@@ -13,6 +17,10 @@ win32 {
     LIBS += -lwinmm
 }
 macx {
+    contains(DEFINES, __UNIX_JACK__) {
+        DEFINES += JACK_HAS_PORT_RENAME
+        LIBS += -ljack
+    }
     DEFINES += __MACOSX_CORE__
     LIBS += -framework CoreMIDI -framework CoreAudio -framework CoreFoundation
 }
--- a/BambooTracker/midi/RtMidi/RtMidi.pri
+++ b/BambooTracker/midi/RtMidi/RtMidi.pri
@@ -5,9 +5,12 @@ HEADERS += \
     $$PWD/RtMidi.hpp
 
 linux {
-    contains(DEFINES, __UNIX_JACK__) {
-        DEFINES += JACK_HAS_PORT_RENAME
+    use_jack {
+        DEFINES += __UNIX_JACK__
         LIBS += -ljack
+        jack_has_rename {
+            DEFINES += JACK_HAS_PORT_RENAME
+        }
     }
     DEFINES += __LINUX_ALSA__
     LIBS += -lasound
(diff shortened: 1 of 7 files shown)


BambooTracker/midi/RtMidi/RtMidi.pri
```diff
@@ -5,6 +5,10 @@ HEADERS += \
     $$PWD/RtMidi.hpp
 
 linux {
+    contains(DEFINES, __UNIX_JACK__) {
+        DEFINES += JACK_HAS_PORT_RENAME
```
OPNA2608, line 9: I believe this only works if we limit what JACK version we support to JACK2. This should pose a problem on macOS where the commonly used JACK version from Homebrew is actually JACK1.

If we really can't detect the JACK version with qmake in the same way RtMidi's build systems do via header inspection or some other way, then maybe we should start thinking about making custom defines for convenience. `-D_USE_JACK1`, `-D_USE_JACK2`, `-D_USE_PULSE` etc.

rerrahkr: I find that [`CONFIG` variable and qmake condition used it](https://doc.qt.io/qt-5/qmake-language.html#configuration-and-scopes).
How about making the `CONFIG` options `use_pulse`, `use_jack1` and `use_jack2` and add each `DEFINES` in these condition?

OPNA2608: That looks like a better idea, yeah.

OPNA2608:
About the build failure on Ubuntu, try apt package `libjack-jackd2-dev` instead of `libjack-dev`.

OPNA2608:
So I actually read abit more into JACK now and it's abit more complicated than JACK1 vs JACK2.

Both JACK1 and JACK2 use the same header files for compilation, the daemons at runtime handle the version stuff on their own.

JACK1 0.125 and JACK2 both have these function declarations in `jack.h`:
https://github.com/jackaudio/headers/blob/eb67fa5ffc915203cdc8757572523c1b036840d9/jack.h#L759-L778
```cpp
int jack_port_set_name (jack_port_t *port, const char *port_name) JACK_OPTIONAL_WEAK_DEPRECATED_EXPORT;
int jack_port_rename (jack_client_t* client, jack_port_t *port, const char *port_name) JACK_OPTIONAL_WEAK_EXPORT;
```
Hence compilation on macOS with `-DJACK_HAS_PORT_RENAME` actually worked!

The reason it initially didn't work on Ubuntu is because Ubuntu 16.04 actually has JACK1 0.12**4**, which hadn't introduced the `jack_port_rename` replacement yet:
https://github.com/jackaudio/headers/blob/a14583ed99d346dacd6579d2d06e273a0c834b24/jack.h#L759-L766
```cpp
int jack_port_set_name (jack_port_t *port, const char *port_name) JACK_OPTIONAL_WEAK_EXPORT;
```
Thus it failed when trying to use the `rename` replacement.
```
midi/RtMidi/RtMidi.cpp: In member function ‘virtual void MidiInJack::setPortName(const string&)’:
midi/RtMidi/RtMidi.cpp:3115:64: error: ‘jack_port_rename’ was not declared in this scope
  jack_port_rename( data->client, data->port, portName.c_str() );
                                                               ^
```

---

So what we actually need is something like `use_jack` and `jack_has_rename`. If `use_jack` is set, setup all the JACK stuff. Additionally if `jack_has_rename` is set, add `DEFINES += JACK_HAS_PORT_RENAME`. Sorry for all the confusion.

OPNA2608:
LGTM, just needs documentation updates. See Readme section `Building`, specifically

- `Dependencies -> Linux / BSD -> Debian / Ubuntu`,
the `libjack-dev` to `libjack-jackd2-dev # for JACK2, or libjack-dev for JACK1`

- `Compilation`,
all the `DEFINES+=` to the corresponding `CONFIG+=` equivalents

OPNA2608:
One last nitpick, explain when dropping `CONFIG+=jack_has_rename` might be needed:

> If you know that building with an old version of JACK that lacks the `jack_port_rename` method or you encounter build problems like
> 
> > error: ‘jack_port_rename’ was not declared in this scope
> 
> , try dropping `CONFIG+=jack_has_rename`.

rerrahkr:
All fixed. I will merge this later.

rerrahkr:
Thanks for your help!

PR #236, opened 2020-07-03
Add workaround for deprecated-copy warning

### LINKED ISSUE #235: Warnings produced after after updating llvm, clang, compiler-rt, libc++, libunwind, lld, lldb and openmp to 10.0.1 rc1 f79cd71e145
FreeBSD recently updated llvm, clang, compiler-rt, libc++, libunwind, lld, lldb and openmp to 10.0.1 rc1 f79cd71e145 (see OSREVISION [1300096](https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/versions.html)) on FreeBSD 13.0-CURRENT.

Since BambooTracker uses `-pedantic` compiler flags, some new warnings cause the compiler to fail. When removing the pedantic compiler flags the tracker compiles just fine but I thought I'd share the build log as you might be interested in fixing warnings at some point:

[typescript-bambootracker-llvmorg-10.0.1-rc2-0-g77d76b71d7d.txt](https://github.com/rerrahkr/BambooTracker/files/4870830/typescript-bambootracker-llvmorg-10.0.1-rc2-0-g77d76b71d7d.txt)

This has definitely no priority.

OPNA2608:
Problematic code is deep in Qt5 Core, utterly unable to fix ourselves.
Warning has been included in Wall/Wextra with GCC 9 / LLVM 10.

(Probably) closes #235. @ehaupt please check if this fixes your compilation problem.

Clang "11" on macOS does not emit this warning and does not understand the `-Wno-deprecated-copy` switch either. *Kinda fishy*, but oh well.

### DIFF
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -42,12 +42,12 @@ CONFIG += c++14
 
 # C/C++ compiler flags
 msvc {
-  CPPFLAGS += /Wall /Wp64 /WX
-  CPPFLAGS += /source-charset:utf-8
+  CPP_WARNING_FLAGS += /Wall /Wp64 /WX
+  CPP_WARNING_FLAGS += /source-charset:utf-8
 }
 else:if(gcc|clang) {
   # Pedantic settings, warning -> error escalation and C standard specification
-  CPPFLAGS += -Wall -Wextra -Werror -pedantic -pedantic-errors
+  CPP_WARNING_FLAGS += -Wall -Wextra -Werror -pedantic -pedantic-errors
   QMAKE_CFLAGS += -std=gnu11
 
   # Attempt to get the compiler version for version-specific handling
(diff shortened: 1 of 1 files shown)


ehaupt:
> (Probably) closes #235. @ehaupt please check if this fixes your compilation problem.

Unfortunately there is still a warning causing it to fail:

[typescript-deprecated-copy-fix-0001.txt](https://github.com/rerrahkr/BambooTracker/files/4872221/typescript-deprecated-copy-fix-0001.txt)

OPNA2608:
> c++ […] -Wno-deprecated-copy […]
> error: definition of implicit copy constructor for 'QByteRef' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]

That's… exactly what I'm trying to downgrade with `-Wno-deprecated-copy`… *Urgh*, quite confusing. :confused: 

https://releases.llvm.org/10.0.0/tools/clang/docs/ReleaseNotes.html

> `-Wextra` now enables `-Wdeprecated-copy`. The warning deprecates move and copy constructors in classes where an explicit destructor is declared. This is for compatibility with GCC 9, and forward looking for a change that’s being considered for C++23. You can disable it with `-Wno-deprecated-copy`.

OPNA2608:
Nevermind, I understand what's happening now.

> `-Wall -Wextra […] -Wno-deprecated-copy […] -Wall -Wextra`

The subsequent `-Wall -Wextra` re-enables the warning. These flags seem to be originating from your Qt5 mkspecs.

On my local Qt5.12.3 machine, `common/gcc-base.conf` (included by GCC and Clang) contains these lines
```qmake
QMAKE_CFLAGS_WARN_ON       += -Wall -W
QMAKE_CFLAGS_WARN_OFF      += -w
```
And my build log does contain `-Wall -W`, among the usual stuff.
> g++ -c -pipe -Wall -Wextra -Werror -pedantic -pedantic-errors -g -std=gnu++1y **-Wall -W** -D_REENTRANT -fPIC -DQT_DEPRECATED_WARNINGS -D__LINUX_PULSE__ -D__LINUX_ALSA__ -D__LINUX_ALSA__ -D__RTMIDI_DUMMY__ -DQT_MULTIMEDIA_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -I. -Ichips -Istream -Iinstrument -Icommand -Imodule -Iio -I/nix/store/1i94w64kxr60mdcajh9lxwrhr1v9vly9-qtmultimedia-5.12.3-dev/include -I/nix/store/1i94w64kxr60mdcajh9lxwrhr1v9vly9-qtmultimedia-5.12.3-dev/include/QtMultimedia -isystem /nix/store/qm0bnz8yccw99xs2ll6h3d9maw1xlw5i-qtbase-5.12.3-dev/include/QtWidgets -isystem /nix/store/qm0bnz8yccw99xs2ll6h3d9maw1xlw5i-qtbase-5.12.3-dev/include/QtGui -isystem /nix/store/qm0bnz8yccw99xs2ll6h3d9maw1xlw5i-qtbase-5.12.3-dev/include/QtNetwork -isystem /nix/store/qm0bnz8yccw99xs2ll6h3d9maw1xlw5i-qtbase-5.12.3-dev/include/QtCore -I. -I/nix/store/p77y5d1ry0lj9qv1ghyrkck8pvcq4451-mesa-19.1.5-dev/include -I. -I/nix/store/jyr3jv2fi2qvl5if43zidivh4w0k4ydn-qt-full-5.12.3/mkspecs/linux-g++ -o RtAudio.o stream/RtAudio/RtAudio.cpp

Qt5.14.2 however has these settings:
```qmake
QMAKE_CFLAGS_WARN_ON       += -Wall -Wextra
QMAKE_CFLAGS_WARN_OFF      += -w
```

I think we could use `QMAKE_CFLAGS_WARN_ON` for our compliance switches, that seems like a much cleaner way to me. Let me do some testing and I'll ping you when I think I got something working.

OPNA2608:
@ehaupt CI passed, how about now?

ehaupt:
Hello @OPNA2608, now there seems to be a syntax error in `BambooTracker.pro`:

```
/usr/ports/audio/bambootracker/work/BambooTracker-31e7213/BambooTracker/BambooTracker.pro:110: Unexpected 'else'.
Error processing project file: /usr/ports/audio/bambootracker/work/BambooTracker-31e7213/BambooTracker/BambooTracker.pro
*** Error code 3
```

OPNA2608:
Rebased on merged BSD fixes, can you try again? There's no `else` in line 110, I'm hoping it was caused by problems on your side with all the PR mixing.

ehaupt:
You're right. I am not sure how this happened. Unfortunately it doesn't quite compile all the way:

[typescript-alsa-jack-jack-has-rename-pulse.txt](https://github.com/rerrahkr/BambooTracker/files/4879361/typescript-alsa-jack-jack-has-rename-pulse.txt)

OPNA2608:
My fault, apologies. The `freebsd` section slipped into the GCC block after rebasing, should be good now?

ehaupt:
Now it compiles like a charm :sweat_smile:

Thank you!

OPNA2608:
Great! :slightly_smiling_face:

About the `freebsd` block, I read in Qt5's mkspecs that there's a `freebsd-g++` target. Is that something worth testing and accounting for or is the FreeBSD (Ports?) default/recommended compiler Clang?

---

Actually, my ALSA headers on Linux have the same zero-length-array declaration and don't appear to throw an error like that when I try building with GCC9, although I might've done something wrong there.

GCC docs indicate to me that [it's considered a GNU C extension](https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html) - I don't know how that plays together with the pedantic switches we use, but I'd expect them to complain about it? I couldn't find a switch directly related to zero-length-array *declarations* either, only for *accessing* such constructs but that's more of a general array bounds issue. ([`-Wzero-length-bounds`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html))

ehaupt:
> About the freebsd block, I read in Qt5's mkspecs that there's a freebsd-g++ target. Is that something worth testing and accounting for or is the FreeBSD (Ports?) default/recommended compiler Clang?

FreeBSD's [tier-1](https://www.freebsd.org/platforms/) archs all use clang by default. Some lower tier's still use gcc. The Ports famework (building packages or installing from source) sets those defaults but is designed to respect [CC/CXX](https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/dads-cc.html)  and [CFLAGS](https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/dads-cflags.html). So technically a tier-1 user could override CC if he prefers so.

The way I understand it, the ALSA packages on FreeBSD are highly adapted for FreeBSD. I think (I could be wrong) they are more of a shim to offer compatibility for FreeBSD because so many applications require ALSA.

OPNA2608:
Then I think this should be ready for a merge. @rerrahkr

PR #230, opened 2020-06-29
Compiler error: redefinition of `__RTMIDI_DUMMY__`

ehaupt:
`__RTMIDI_DUMMY__` gets defined in `BambooTracker/midi/RtMidi/RtMidi.pri` and `BambooTracker/midi/RtMidi/RtMidi.cpp` causing the following compiler warning:

```
/usr/local/lib/qt5/bin/moc -DQT_DEPRECATED_WARNINGS -D__RTMIDI_DUMMY__ -DQT_NO_DEBUG -DQT_MULTIMEDIA_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB --include /usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/moc_predefs.h -I/usr/local/lib/qt5/mkspecs/freebsd-clang -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/chips -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/stream -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/instrument -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/command -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/module -I/usr/ports/audio/bambootracker/work/BambooTracker-0.4.3/BambooTracker/io -I/usr/local/include/qt5 -I/usr/local/include/qt5/QtMultimedia -I/usr/local/include/qt5/QtWidgets -I/usr/local/include/qt5/QtGui -I/usr/local/include/qt5/QtNetwork -I/usr/local/include/qt5/QtCore -I/usr/include/c++/v1 -I/usr/lib/clang/8.0.1/include -I/usr/include gui/instrument_selection_dialog.hpp -o moc_instrument_selection_dialog.cpp
--- RtMidi.o ---
midi/RtMidi/RtMidi.cpp:60:11: error: '__RTMIDI_DUMMY__' macro redefined [-Werror,-Wmacro-redefined]
  #define __RTMIDI_DUMMY__
          ^
<command line>:2:9: note: previous definition is here
#define __RTMIDI_DUMMY__ 1
        ^
--- moc_s98_export_settings_dialog.cpp ---
[...]

PR #238, opened 2020-07-04
Fix audio and MIDI configuration

### LINKED ISSUE #240: [macOS] Can't click MIDI device option
macOS 10.15, a user is unable to _click_ on any MIDI device option. It gives them an error sfx and doesn't do anything. They _can_ however use the arrow keys and enter to select their device of choice, which otherwise works as expected.

grafik

rerrahkr:
#230
- Change audio device list in the configuration dialog when the selected API is changed
- Add MIDI API settings
- Update RtMidi to the latest version 4.0.0

EDIT: After this PR is merged, close #240.


OPNA2608:
Tested PulseAudio & ALSA audio as well as ALSA & Dummy MIDI on Linux, quite works well from what I can tell. Can't test JACK or OSS, maybe (since it's related to #230) @ehaupt could help out?

Personally I think it'd look better to have the audio output API combobox together with the rest of the audio settings like in the MIDI group, and the MIDI input device as a combobox like the audio output devices. Something like this (mockup in Paint):

grafik

OPNA2608:
Configuration dialogue prints

> QMetaObject::connectSlotsByName: No matching signal for on_soundAPIComboBox_currentIndexChanged(QString)

in Terminal, Audio Device combobox doesn't update anymore.

rerrahkr:
Oops, sorry I forgot to reconnect them after renaming.

ehaupt:
> Tested PulseAudio & ALSA audio as well as ALSA & Dummy MIDI on Linux, quite works well from what I can tell. Can't test JACK or OSS, maybe (since it's related to #230) @ehaupt could help out?

I tried  4d68421 but unfortunately the branch was created prior the (for me) necessary changes made in #230 (unmerged). It gets a bit difficult to track all the moving parts.

OPNA2608:
Afew problems that I'm noticing with this:

1. If ALSA has been used & saved and BambooTracker is started while the last used device is busy (ALSA applications block the whole device they're using), BambooTracker crashes with a SIGSEGV. The issue is perhaps very similar to the MIDI initialization crash from #172, just encountered on the audio side this time.
<details>
<summary>Logs</summary>

```
[nix-shell:~/devel/BambooTracker/BambooTracker/BambooTracker]$ ./BambooTracker 
Translation "qt_de_DE" from "/nix/store/0fm4l6wsbg0vh2q2mah811bw9wy2ykpy-qtbase-5.12.3/translations"
Translation "bamboo_tracker_de_DE" from "/home/bt1cn/devel/BambooTracker/BambooTracker/BambooTracker/.qm"
Using emulator: Nuked OPN-Mod

RtApiAlsa::getDeviceInfo: snd_pcm_open error for device (hw:0,0), Device or resource busy.


RtApiAlsa::getDeviceInfo: snd_pcm_open error for device (hw:0,0), Device or resource busy.


RtApiAlsa::probeDeviceOpen: pcm device (hw:0,0) won't open for output.

Segmentation fault (Core dumped)
```
</details>

---

2. When the startup does work, the ALSA API is seemingly initialized wrong. If i used the "default" device (bottom-most entry) with 55466Hz sample rate and attempt to relaunch BambooTracker and play a track, it sounds like it's generating 55466Hz audio but playing it back at 48000Hz by slowing it down. Additionally, the device box now says that a completely different device is in use than the last time (top-most entry in the list).

I think it might be reconnecting to PulseAudio via ALSA somehow? Looks to me like the startup audio initialization is still using the previous system (I noticed the same playback phenomena afew days ago).

<details>
<summary>Here's some information from the BambooTracker.ini file:</summary>

- After using ALSA successfully and closing.
  > [Sound]
bufferLength=20
emulator=1
realChipInterface=0
sampleRate=55466
soundAPI=ALSA
soundDevice=default

- After restarting, experiencing the aforementioned problems and closing without changing any settings.
  > [Sound]
bufferLength=20
emulator=1
realChipInterface=0
sampleRate=55466
soundAPI=ALSA
soundDevice=PulseAudio
</details>

---

3. The PulseAudio API is not displaying the available sinks (output targets in PulseAudio), only "PulseAudio". I tested this by creating a null-sink (connected to nothing) and it continued showing only "PulseAudio". Maybe RtAudio is not capable of picking up the actual sinks?

grafik

Example from Discord properly picking up the 2 names sinks:
grafik

OPNA2608:
@ehaupt Right, apologies. Let's wait for that PR to get merged and the other PRs rebased if necessary before asking for any more feedback from the BSD side.

ehaupt:
@OPNA2608 no worries. I like the idea of being able to set the MIDI device on FreeBSD to dummy as MIDI support is non existent without something like [timidity](https://en.wikipedia.org/wiki/TiMidity%2B%2B). Setting the MIDI device to 'Dummy device' should then quiesce the MIDI warning and I could set the output device to pulseaudio, OSS or ALSA.

OPNA2608:
> Maybe RtAudio is not capable of picking up the actual sinks?

I went back to `master`'s state while testing something and noticed that the old behaviour actually used the proper PulseAudio sink names, albeit for all APIs.

grafik

OPNA2608:
Could you rebase this on current master? With #236 merged there should be nothing in the way of getting feedback from the BSD side, hoping this might help fix OSS and give some insight on how well the JACK side is working.

rerrahkr:
> If ALSA has been used & saved and BambooTracker is started while the last used device is busy (ALSA applications block the whole device they're using), BambooTracker crashes with a SIGSEGV.

It seems that an error is occurred in `RtAudio::openStream` called from `AudioStreamRtAudio::initialize`, but I use try-catch statement:
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/stream/audio_stream_rtaudio.cpp#L51-L58
Before crash, is the dialog warning `Could not open the audio stream. Please change the sound settings in Configuration.` shown?

> the device box now says that a completely different device is in use than the last time (top-most entry in the list).

A similar phenomenon occurs when DirectSound is set on my Windows 10. After saving and closing with "Default Device" and starting, it will be changed to "Line" (my audio interface). When I use the debugger, RtAudio doesn't seem to find the "Default Device" at startup. If there is no device stored in the settings, the tracker selects the first device ("Line").
I think the sample rate problem probably occurs when `RtAudio::openStream` doesn't get the desired rate. I find `unsigned int getStreamSampleRate( void );` exists, so I'll try to fix it to make the appropriate rate changes.

> > Maybe RtAudio is not capable of picking up the actual sinks?
> 
> I went back to `master`'s state while testing something and noticed that the old behaviour actually used the proper PulseAudio sink names, albeit for all APIs.

`master` still uses `QAudioDeviceInfo::availableDevices(QAudio::AudioOutput)`. Perhaps RtAudio may not be able to get the Sink name, but I'll do a little more research.

OPNA2608:
> Before crash, is the dialog warning `Could not open the audio stream. Please change the sound settings in Configuration.` shown?

It wasn't, the application immediately terminated without any GUI components. After the rebase however I'm unable to replicate this crash.

> After saving and closing with "Default Device" and starting, it will be changed to "Line" (my audio interface). When I use the debugger, RtAudio doesn't seem to find the "Default Device" at startup. If there is no device stored in the settings, the tracker selects the first device ("Line").
> […]
> I think the sample rate problem probably occurs when `RtAudio::openStream` doesn't get the desired rate. I find `unsigned int getStreamSampleRate( void );` exists, so I'll try to fix it to make the appropriate rate changes.

The sample rate problem is really a red herring, a bug resulting from the audio device bug. If I use a debugger and break at the very first `AudioStreamRtAudio::initialize` after restarting and having last used ALSA, I get

```
Thread 1 "BambooTracker" hit Breakpoint 1, AudioStreamRtAudio::initialize (this=this@entry=0x35c7c80, rate=55466, 
    duration=duration@entry=20, intrRate=intrRate@entry=60, backend=..., device=...) at stream/audio_stream_rtaudio.cpp:17
17	{
(gdb) printqs5dynamic backend
(Qt5 QString)0xffff5170 length=4: "ALSA"
(gdb) printqs5dynamic device
(Qt5 QString)0xffff5180 length=10: "PulseAudio"
```

ALSA _shouldn't_ connect to the PulseAudio device to begin with, that's just asking for trouble. This is actually the current behaviour on master and you can reproduce this very bug there by picking 55466Hz, ALSA and any device (which are PulseAudio sinks by default). If you attempt to properly connect to a PulseAudio sink via the PulseAudio API, the API actually responds with

```
RtApiPulse::probeDeviceOpen: unsupported sample rate.
```

I don't think this is really worth investigating. Since it's an API<->device problem, fixing the device detection bug so it doesn't default to the wrong API's device should resolve this automatically.

OPNA2608:
The following line creates the wrapper around RtAudio on startup.
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/gui/mainwindow.cpp#L625

The constructor initializes RtAudio with the `UNSPECIFIED` API
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/stream/audio_stream_rtaudio.cpp#L6-L10
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/stream/RtAudio/RtAudio.hpp#L439

which iterates over the compiled-in APIs until it finds an API with at least 1 device, which happens to be PulseAudio in my case.
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/stream/RtAudio/RtAudio.cpp#L131-L137
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/stream/RtAudio/RtAudio.cpp#L260-L269

With the Pulse API selected, we then try to find the saved _ALSA_ device via API device listing, which obviously fails.
https://github.com/rerrahkr/BambooTracker/blob/2bbae36d051601b03edb5dd39dedf79f3cd9e443/BambooTracker/gui/mainwindow.cpp#L635-L641

The solution for this should be to initialize RtAudio with the saved API.

rerrahkr:
> After saving and closing with "Default Device" and starting, it will be changed to "Line" (my audio interface). When I use the debugger, RtAudio doesn't seem to find the "Default Device" at startup. If there is no device stored in the settings, the tracker selects the first device ("Line").

This is because the tracker got devices of WASAPI, the problem you pointed out was happening.

To avoid this, I added API check and fixed device check to consider API.

OPNA2608:
> To avoid this, I added API check and fixed device check to consider API.

Confirmed fixed, that just leaves the PulseAudio sinks problem.

2 more questions: Would it be possible to have the error output of the audio API initialization in the error dialogue? As of right now, if there are any errors on the API side all it tells you is "change something". Meanwhile the terminal contains the relevant output to understand what went wrong and why it went wrong. This information is completely unavailable on the release builds on Windows (needs `CONFIG+=console` like the AppVeyor debug builds to get stdout) and abit out of the way on macOS (need to run the binary inside the Appbundle instead of BambooTracker.app itself). Disregard this idea if it'd be too much effort, it seems to work fine for most users so far.

grafik

Secondly is there a strong reason for keeping the MIDI device textbox + `…` button? I think it looks abit out of place, with everything else being an auto-updating combobox. The device name in the textbox also lingers around when selecting a different API.

Otherwise LGTM.

@ehaupt Maybe you could give this a shot now & see if this fixes OSS and JACK for you?

rerrahkr:
> 2 more questions

I fixed them. I couldn't check API initialize error dialog, does it works well when using PulseAudio?

I searched how to get sink names of PulseAudio, however, at least, unfortunately it seems RtAudio doesn't prepare its method in the latest release.

OPNA2608:
> I couldn't check API initialize error dialog, does it works well when using PulseAudio?

Looks good to me.

grafik

One last thing, could the comboboxes be set as disabled if no devices were found for the specified API? Current behaviour leaves the comboboxes without any entries if the corresponding API couldn't be contacted at runtime (no running JACK server) or the API provides no devices in general (MIDI dummy API).

grafik

---

> I searched how to get sink names of PulseAudio, however, at least, unfortunately it seems RtAudio doesn't prepare its method in the latest release.

This seems to be a known bug upstream, see https://github.com/thestk/rtaudio/issues/240. We could try bumping the RtAudio version to the head of master or we wait until eventually a new version is released.

OPNA2608:
LGTM. Don't know why CI errored, merges just fine on my end.

ehaupt:
> @ehaupt Maybe you could give this a shot now & see if this fixes OSS and JACK for you?

I've just tried to compile it with jack, oss, and pulse enabled. Unfortunately there is a compiler warning causing the build to fail due to the `-pedantic` flags:

[buildlog_bambootracker_alsa_jack_pulseaudio.txt](https://github.com/rerrahkr/BambooTracker/files/4936045/buildlog_bambootracker_alsa_jack_pulseaudio.txt)

Removing the `-pedantic` flags it builds (unsurprisingly) fine.

Starting the tracker I get:

image

followed by:

image

Meanwhile in the terminal where I launched the tracker:

```
Translation "qt_C" from "/usr/local/share/qt5/translations"
Translation "bamboo_tracker_C" from "/usr/local/bin/../share/BambooTracker/lang"
Using emulator: MAME YM2608
ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: No such file or directory

MidiInAlsa::initialize: error creating ALSA sequencer client object.

Cannot initialize MIDI In.

MidiInAlsa::initialize: error creating ALSA sequencer client object.


MidiInDummy: This class provides no functionality.

ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: No such file or directory

MidiOutAlsa::initialize: error creating ALSA sequencer client object.

Cannot initialize MIDI Out.

MidiOutAlsa::initialize: error creating ALSA sequencer client object.


MidiOutDummy: This class provides no functionality.


RtAudio: no compiled support for specified API argument!


RtApi::openStream: output device parameter value is invalid.

QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder
QWidget::setWindowModified: The window title does not contain a '[*]' placeholder

```

Changing to OSS i get:

image

Plus at the terminal:

```
RtApiOss::getDeviceInfo: device (pcm0:play:dsp0.p0) data format not supported by RtAudio.
RtApiOss::probeDeviceOpen: device (pcm0:play:dsp0.p0) is busy.
```
Switching to jack:

image

Console:

```
RtApi::openStream: output device parameter value is invalid.
```

Jackd is running with:

```
jackd 0.125.0rc1 (COOKEDMODE=on)
Copyright 2001-2009 Paul Davis, Stephane Letz, Jack O'Quinn, Torben Hohn and others.
jackd comes with ABSOLUTELY NO WARRANTY
This is free software, and you are welcome to redistribute it
under certain conditions; see the file COPYING for details

JACK compiled with System V SHM support.
loading driver ..
oss_driver: /dev/dsp : 0x10/2/44100 (4096)
oss_driver: indevbuf 4096 B, outdevbuf 4096 B
oss_driver: not using barrier mode, (single thread)
```

So far pulseaudio remains the only working API on FreeBSD.

OPNA2608:
> Unfortunately there is a compiler warning causing the build to fail

Introduced by #239, will open a fix PR shortly. This should work:

```diff
diff --git a/BambooTracker/BambooTracker.pro b/BambooTracker/BambooTracker.pro
index a01d187..580c606 100644
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -90,7 +90,7 @@ else:clang|if(gcc:!intel_icc) {
     # macOS 10.14 (LLVM 11.0.0) targeting gnu++1y (C++14) errors when
     # using system-installed JACK headers in RtAudio & RtMidi
     # /usr/local/Cellar/jack/0.125.0_4/include/jack/types.h:(389,411)
-    clang-apple {
+    use_jack {
       CPP_WARNING_FLAGS += -Wno-deprecated-register
     }   
 
```

> JACK
> `RtApi::openStream: output device parameter value is invalid.`

When you pick JACK for your audio output does the device combobox list any (JACK-typical) devices or is it disabled (same for OSS)? Additionally you could check if selecting JACK as your MIDI API shows any ports aside from "Virtual port" or if it prints

```

MidiInJack::initialize: JACK server not running?


MidiInJack::initialize: JACK server not running?

```

on the terminal?

> pulseaudio remains the only working API

Just an idea: Are you still running Pulseaudio in the background while attempting OSS or JACK? I don't know your exact audio setup or how it's handled on FreeBSD, but on Linux running JACK and PulseAudio at the same time without a very special configuration of either of them usually causes one of the sound servers to block the other from working.

Have you tried killing PulseAudio before launching JACK / trying OSS? What sample rates have you tried selecting, JACK seems to connect fine to `/dev/dsp` with 44100Hz? Are other applications capable of connecting to JACK when launched this way or is it just BT that's unable to detect it?

rerrahkr:
There is no class having dependency to Qt multimedia, so I remove `multimedia` from project file.

rerrahkr:
This PR mainly fixes API selections. The fix also affects all systems, but the error seems to have occurred only on FreeBSD. It may have problems outside the scope of this fix.
How about merging this PR and opening another issue to discuss errors in FreeBSD? Or should I keep this PR opening until all the problems are resolved?

OPNA2608:
This PR has abunch of good stuff in it, it's been open for roundabout 2mo now, the mentioned build problem might've been fixed by #241(?) and afaict there aren't any new regressions in functionality introduced by this PR - just already existing problems on FreeBSD.

I'm in favour of merging this PR at this point; we should track our progress at getting the other audio APIs working on FreeBSD in a proper issue. :+1:

rerrahkr:
I opened #265 for discussing BSD problems, so I'll merge this PR after CI checks. Thanks.

PR #239, opened 2020-07-08
Improve compiler & version detection

OPNA2608:
~~Depends on #236, don't merge yet.~~ Ready for merge.

I've dug abit more around the Qt5 mkspecs and found handy macros for cleanly identifying the configured compiler version, *phew*. (`mkspecs/features/data/macros.cpp` and `mkspecs/features/qt_common.prf`)

I added some more `CONFIG+=`s like for the audio API options to differentiate regular LLVM from Apple's LLVM (see below), might rethink that after abit more sleeping & thinking.

Turns out that Apple's Clang actually uses a custom versioning scheme that deviated from upstream LLVM's version releases. More information about this can be found here: https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_11.x_(since_Free_On-Device_Development)

I decided to drop the version limitation for the JACK on macOS workaround; The older compiler on CI understands the switch just fine even if it doesn't need it. For the deprecated-copy one from #236, there's no Apple LLVM release yet that's affected by this error so I limited it to regular Clang only for now as noted in the comment.

```
# TODO: Extend to clang-apple once LLVM 10 reaches the platform and they decide on a new custom version
```

MSVC part went completely untested as usual.

### DIFF
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -42,26 +42,40 @@ CONFIG += c++14
 
 # C/C++ compiler flags
 msvc {
+  message("Qt configured with MSVC")
+  message("Compiler is version" $$QT_MSC_FULL_VER)
   CPP_WARNING_FLAGS += /Wall /Wp64 /WX
   CPP_WARNING_FLAGS += /source-charset:utf-8
 }
-else:if(gcc|clang) {
+else:clang|if(gcc:!intel_icc) {
   # Pedantic settings, warning -> error escalation and C standard specification
   CPP_WARNING_FLAGS += -Wall -Wextra -Werror -pedantic -pedantic-errors
   QMAKE_CFLAGS += -std=gnu11
 
-  # Attempt to get the compiler version for version-specific handling
-  # Clang returns an ancient pinned GCC version, attempt hacky & ugly output grepping instead
+  # Get the compiler version for version-specific handling
   clang {
-    message("Qt configured with Clang")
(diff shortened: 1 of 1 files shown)


BambooTracker/BambooTracker.pro
```diff
@@ -42,55 +42,89 @@ CONFIG += c++14
...
 # C/C++ compiler flags
 msvc {
-  CPPFLAGS += /Wall /Wp64 /WX
-  CPPFLAGS += /source-charset:utf-8
+  message("Qt configured with MSVC")
+  message("Compiler is version" $$QT_MSC_FULL_VER)
+  CPP_WARNING_FLAGS += /Wall /Wp64 /WX
+  CPP_WARNING_FLAGS += /source-charset:utf-8
 }
 else:if(gcc|clang) {
```
OPNA2608, line 50: This `gcc|clang` is technically meaningless. `clang` is handled as a derivate of and additionally defines `gcc` so `gcc` would be enough. (Same technically applies for Intel's `icc` compiler -`intel_icc` option - but we don't test for it and I have no way to test it anyway)

I think it's better to keep this in some form for a quicker overview of what compilers we expect and support though. maybe `clang:!intel_icc:gcc` would be better in that we explicitly exclude `icc`?

rerrahkr: I think it's better to keep it for clarity.

> `clang|if(gcc:!intel_icc)`

I don't know how many users are using icc (it's new to me), but I agree with that, in that we should explicitly indicate that we know that it and gcc are different.

rerrahkr:
Thanks for your contribution!

PR #254, opened 2020-08-09
CI: Linux AppImage

OPNA2608:
Title says it all. Also includes minor changes to the Windows CI.

Just a draft because I still need to sort out some missing libraries in the final AppImage, I also need to clean up several minor things about this. It should already run on most glibc systems though.

I haven't looked into possible size optimizations yet, it's on my to-do list. I do believe though that linuxdeployqt runs what's essentially several iterations of `ldd` on the binary to add most libraries that are being references as required that way. It'll prolly turn out to be a very big archive regardless though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,16 +13,17 @@ jobs:
         - QTDIR=$HOME/Qt/win7
         - QTVER=5.14.1
         - QTTCH=win32_mingw73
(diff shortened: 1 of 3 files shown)


OPNA2608:
Gonna close this. I have some other plans for Linux releases amidst other CI improvements, just need to finish some things and tidy up the commit history before I open the PR.

PR #268, opened 2020-10-20
GitHub: Disable blank issues

OPNA2608:
Should hopefully encourage more ppl to use the *"bug report"* template for their *bug reports*. ;(

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false


rerrahkr:
If only templates can be selected, is it better to prepare Question templates as well?

OPNA2608:
I don't think questions belong into the repository: it's a code repository & issue tracker, not a support forum. The one-off question is gonna be unavoidable regardless ofc - you can always just select any template and erase all of its content, they're merely suggestions for reporters. I'm kinda fearing that adding an explicit question template could invite very generic questions and make the issue tracker abit messier as a result.

If you don't mind me spit-balling for a second here, I think some community chat that we could link in the README would be more suited for responding to questions. Reoccurring questions could then propagate into a F.A.Q. section on the wiki, and it could be a good place to collect BambooTracker users, let them share tricks, modules and skins, ask for direct feedback on decisions etc. [I know you're using your Twitter for very similar purposes so far](https://twitter.com/RerrahKRynn/status/1167809821451538432) but I'm not a huge fan of Twitter myself. Some options for such a community thing could be a [Discord](https://discord.com/) server, a [Matrix](https://matrix.org/) [room](https://element.io/) or an IRC channel. I would know some ppl I'd invite straight-away as well.

Let me know what you think, I can always just add a question template or drop the PR altogether if you'd prefer that.

rerrahkr:
I understand why we shouldn't have a question template.

Regarding a community chat, I think BambooTracker should also have a forum like FamiTracker and Deflemask. I also would like to create a place where we can archive or post modules and instruments and share them. However, it is not possible to prepare them immediately because it is difficult for me to manage the server.
Therefore, I agree with preparing some community chat as a place for simple communication between users. I'm a beginner of Discord, but I know that I can ask questions, discuss, and publish modules and instruments for each channel. Also, it seems that there is a Discord server in Klystrack etc.

I can merge this PR, but leave it as it is until we have a place to ask questions.

OPNA2608:
I went ahead and set up a "Proof of Concept" Discord server, let me know what you think about it: https://discord.gg/gBscTMF

rerrahkr:
Discord server is opened, so I merge this PR. Thanks for your contributions!

PR #271, opened 2020-10-23
More build/release information in README, bump versions on CI, Linux releases

OPNA2608:
Draft: Commit history still needs to be cleaned up, it's a mess due to plenty of CI debugging. Not done with the README either: I like the extra information but it feels to formal & verbose and manual build instructions still need to be touched up - Windows specifically. And the added branch & my CI release information in the Travis config still need to be reverted ofc.

Goal / Changes of this PR:

- Expand on our build & release-specific details. Supported & tested operating systems & toolchains, when releases are being made (whenever it feels appropriate I think?)
- Further improve CI:
  - Nicer style for some environment variables
  - `windows` bumped to latest Qt5.15 release. Required a change in `ci/package_windows.sh` - `lrelease` now works according to how Qt *thinks* it should work, placing translations in `release/*.qm` (and presumably `debug/*.qm` if not building with `CONFIG+=release`?, untested)
  - `macos` bumped to latest macOS & Xcode version supported by Homebrew's Qt5.15 version
    (`xcode12.2` builds but prints a warning about untested & too new SDK version)
  - `macos-legacy` *drastically* bumped down to be actually-legacy, by pinning an old commit of Homebrew with Qt5.6.2 and using Travis CI's oldest available macOS image. Build succeeds and should support macOS 10.7 to 10.11 according to [Qt5 docs](https://doc.qt.io/archives/qt-5.6/supported-platforms.html) of the time, but I know absolutely nobody who could test that claim.
    This sadly leaves 10.12 with supposedly no supported build, but I'd be surprised if 10.11 really is a hard cut-off for `macos-legacy`. It was tested on 10.14, but failed there.
- Implemented a new builder on Travis CI. Using Ubuntu as a base, it installs the Nix package manager with the cutting-edge package channel and builds my upstream package definition for BambooTracker (doubling as a packaging test for me :stuck_out_tongue:), with an override to use the local checkout as the source.
- When building a git tag, a [nix-bundle](https://github.com/matthewbauer/nix-bundle) is generated & uploaded as our Linux release. Unlike the very similar AppImage, which while more common [doesn't work on all distributions, requiring the GNU libc implementation](https://www.reddit.com/r/voidlinux/comments/favmwm/appimage_and_musl/), a nix-bundle includes all of its dependencies (much bigger as a result) and only needs a kernel with support for user namespaces.
I might improve this abit more, by dropping all the misc files from the Nix build with more overrides and re-adding them during the pre-deploy step by tarballing the bundle + all the docs & examples, similar to how the macOS .app bundles are being delivered.

[Test-releases can be found here](https://github.com/OPNA2608/BambooTracker/releases/tag/v0.5.0). Would appreciate any possible feedback, especially on `macos-legacy` and `linux`. Also would love to see the latter tested on any more x86_64-linux systems (tested on Ubuntu 18.04 and Void Linux).


freq-mod:
About feedback on linux, this thing crashed on my system when browsing through filesystem:
`**
GLib-GIO:ERROR:../gio/gunixvolume.c:324:eject_mount_do: assertion failed (error == NULL): Failed to execute child process ?mount? (No such file or directory) (g-exec-error-quark, 8)
Bail out! GLib-GIO:ERROR:../gio/gunixvolume.c:324:eject_mount_do: assertion failed (error == NULL): Failed to execute child process ?mount? (No such file or directory) (g-exec-error-quark, 8)`

OPNA2608:
Can I get more details about your distribution? When does it occur, immediately when opening the file browser without moving or navigating to some directory? What directory you were in and which directory were you trying to access?

The error sounds like a mounting-related problem. For security reasons, the bundle [does some filesystem mapping and only allows very specific directories](https://github.com/matthewbauer/nix-bundle/blob/83742ac8ea4230ca239f948a6b2127449b25ece8/nix-user-chroot/main.cpp#L113-L124). When trying to access any unmounted directory (in the chroot mapping, but mounted on my actual system), i get booted back to `/` by the file manager.

Maybe something about that caused the error on your end?

freq-mod:
Oh sorry. Manjaro x64, Xfce 4.14, kernel 5.9.1. I think it happened after I navigated to run, ten to root. I will try to replicate this crash.

freq-mod:
also when sound API is set to pulse you can't set the sample rate to 55466 Hz, it throws an error about unsupported sample rate (44 and 48 kHz are fine). Then I set the device to ALSA but sound became distorted (slower and down few semitones in pitch). 

About the crash, I cant replicate it rn

OPNA2608:
> also when sound API is set to pulse you can't set the sample rate to 55466 Hz, it throws an error about unsupported sample rate (44 and 48 kHz are fine).

This was introduced by some fixes in #238. The way we talked to the system's audio APIs and opened audio streams to API devices used to be completely incorrect. With this "fixed", using PA with 55466Hz appears to be unsupported. I'll try to look into whether that's actually a bug on our side or just PA/RtAudio genuinely not supporting this, for whatever reason. In the latter case the best we could do is dynamically downsample 55466Hz to 48000Hz.

> Then I have set the API to ALSA but sound became distorted (slower and down few semitones in pitch).

I think I remember this being mostly fixed by the above PR as well? I remember very consistently reproducing this awhile ago, but some commit on master had fixed this before I could report it. It's possible your current configuration in `~/.config/BambooTracker/BambooTracker.ini` might be causing this? I'll see if I can replicate this later.

> Jack doesn't work at all, output parameter device is invalid (I can't see any)

[JACK](https://github.com/jackaudio/jack2) is a low-latency sound API that requires a running JACK server on your system. Without one installed there will be no possible devices found, so it can't be initialized either. In these "fat" prebuilt binaries we want to provide as many options as possible because we have absolutely *no* idea what system a user might be running it on, so this is fine.

OPNA2608:
> I'll see if I can replicate this later.

Very reliably so, and dead-easy too. I'll open a tracking issue for it because this PR has nothing to do with that, but I can't tell you yet when I'll get to debug this.

alexmyczko:
you are aware of https://repology.org/project/bambootracker/versions ?

oh and: https://buildd.debian.org/status/package.php?p=bambootracker&suite=sid

OPNA2608:
Closed by #285 because a bulk of the changes here don't matter anymore after switching to GitHub Actions. I'll open another PR with just documentation improvements some time later.

---

You mean regarding this line?

> If your package manager's version of BambooTracker is missing from the macOS / Linux section and you want to add it to the list, please open a pull request or tell us about it on our Discord server!

I could link the repology overview in the README as well, but I'd prefer to list actual install instructions (either on the README or wiki) for packages that I know are configured well.

For example I know there's an AUR package that sorta auto-updates itself (I don't use Arch but that's what it looks like), but the expression hasn't been touch in a long while so it lacks optional compile-time dependencies for a greater possible audio API selection at runtime (which I *would* recommend since they only require the libraries and won't hinder ALSA-only users much), includes outdated workarounds, manually installs some files that should be covered by the install target now, requires a dependency we dropped etc… :confused:

PR #274, opened 2020-10-25
Audio stream initialization by multiple APIs

### LINKED ISSUE #270: [MacOS] - Incorrect Default Audio API
## Bug Description

Audio defaults to JACK, even on MacOS systems not using JACK.

## How to reproduce

On a macOS system without a BambooTracker.ini, and a test release based on current master, BT picks JACK as the default on startup and an error window about audio initialization appears. No JACK device could be detected so initialization fails. if no previous API and/or device has been saved, like on a 1st time launch, BT should silently try the APIs in the preferred order. e.g. JACK -> no devices -> CoreAudio -> first/primary device.

## System Information
- **Operating System**: MacOS Mojave 10.14.5
- **BambooTracker Version**: `unstable-1c9a1ceb8339f6d132e5d7faef775cbc3ca269f8`
- **Build Type**: on travis CI, @OPNA2608's release testing

rerrahkr:
Close #270

I cannot test in my macOS, which is borrowed from my friend, because I cannot install JACK to it.

@N-SPC700 Could you please check that this works well?

@OPNA2608 This PR will only try to initialize with multiple APIs the first time it starts, strictly speaking, when no APIs are specified in the configuration file. The available API is automatically selected at the first startup, but if an error like  [your report](https://github.com/rerrahkr/BambooTracker/issues/270#issuecomment-713639218) occurs at the second and subsequent launches, it is considered that the cause is the user's operation, so the tracker will suggest that the user manually change the API settings.

### DIFF
--- a/BambooTracker/gui/mainwindow.cpp
+++ b/BambooTracker/gui/mainwindow.cpp
@@ -659,36 +659,53 @@ MainWindow::MainWindow(std::weak_ptr<Configuration> config, QString filePath, QW
 	}, bt_.get());
 	QObject::connect(stream_.get(), &AudioStream::streamInterrupted, this, &MainWindow::onNewTickSignaled);
 	QString audioApi = utf8ToQString(config.lock()->getSoundAPI());
-	bool savedApiExists = false;
-	const std::vector<QString> audioApis = stream_->getAvailableBackends();
-	for (const QString& api : audioApis) {
-		if (api.toUtf8().toStdString() == config.lock()->getSoundAPI()) {
-			savedApiExists = true;
-			break;
+	if (audioApi.isEmpty()) {	// On the first launch
+		bool streamState = false;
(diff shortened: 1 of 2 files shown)


OPNA2608 (approved):
Tested with no configuration file, successfully & silently falling through from JACK -> PA, audio works out-of-the-box.

```
RtApiJack::getDeviceInfo: Jack server not found or connection error!


RtApi::openStream: output device parameter value is invalid.

RtAudio pulse: _NOT_ running realtime scheduling
```

Manually setting JACK and restarting still errors as expected.

N-SPC700:
Using the latest build from appveyor (`e690e50a`), I can confirm this works, being able to start the program without changing audio settings.

rerrahkr:
Thanks testing, so it seems okay. I'll merge this.

PR #280, opened 2020-11-01
Synchronize sample rate between the emulation and the audio stream

### LINKED ISSUE #279: [Linux] - High sample rate on ALSA causes slow & downpitched audio
> #### Checklist
> 
> 
> 

---

## Bug Description
Selecting the ALSA API with a sample rate of 55466Hz properly connects to the ALSA backend of the system, but plays the 55466Hz audio stream as if it was a 48000Hz one. The effect of this is that whatever is being generated & played back sounds slower & lower pitched than it's supposed to.

I'm suspecting that this might be caused by a fixed sample rate somewhere in the `emulated chip` -> `ALSA backend` chain, so 44100Hz would have the opposite effect of being faster & pitched up, though by such a tiny percentage that it's hardly noticable.

See also https://github.com/rerrahkr/BambooTracker/pull/271#issuecomment-715927147

> Then I have set the API to ALSA [with 55466Hz sample rate] but sound became distorted (slower and down few semitones in pitch).

## How to reproduce

1. Start BambooTracker
2. Go to `Configuration` -> `Sound`
3. Select ALSA API, any device that works and isn't occupied
4. Choose 55466Hz sample rate
5. `OK`
6. Load & play a module

## System Information

- **Operating System**: NixOS 20.03 (Markhor) x86_64
- **BambooTracker Version**: master 1c9a1ceb8339f6d132e5d7faef775cbc3ca269f8
- **Build Type**: any

rerrahkr:
close #279 

Check the sample rate between the configuration and ALSA.
If they are different, the emulation sample rate is changed and show warning.

@OPNA2608  Could you please check this PR works well? I cannot test about ALSA on Windows.

### DIFF
--- a/BambooTracker/gui/mainwindow.cpp
+++ b/BambooTracker/gui/mainwindow.cpp
@@ -673,7 +673,16 @@ MainWindow::MainWindow(std::weak_ptr<Configuration> config, QString filePath, QW
 							  bt_->getModuleTickFrequency(), audioApi, audioDevice, &streamErr);
 			if (streamState) break;
 		}
-		if (!streamState) showStreamFailedDialog(streamErr);
+		if (streamState) {
+			uint32_t sr = stream_->getStreamRate();
+			if (config.lock()->getSampleRate() != sr) {
+				showStreamRateWarningDialog(sr);
+				bt_->setStreamRate(sr);
+			}
+		}
+		else {
+			showStreamFailedDialog(streamErr);
+		}
 	}
 	else {	// Ordinary launch
 		bool savedApiExists = false;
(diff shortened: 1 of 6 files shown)


OPNA2608:
The warning message works and the pitch has been fixed now, but the playback speed is still too slow. It's hard to record ALSA without fiddling with the settings or using a hardware-level loopback cable, but it sounds like if you recorded some audio at 48000Hz and used Audacity's "change tempo" option to stretch it to (55466/48000) percent while preserving pitch.

rerrahkr:
Sorry, I forgot to change the tick interval calculation. I force-pushed a new commit, so I wish it solves this issue.

OPNA2608 (approved):
Sounds fine to me now. :+1:

rerrahkr:
Thanks for testing!

PR #283, opened 2020-11-03
Change the position of the QApplication instantiation

### LINKED ISSUE #282: [macOS] - "Fresh" run tries to open nonsense file
> #### Checklist
> 
> 
> 

---

Reporting for @N-SPC700

## Bug Description

A "fresh" first run (`mv ~/.config/BambooTracker ~/.config/BambooTracker-bak`) of BambooTracker errors about trying to open a nonsense module called `-psn_0_...`

grafik

I did some research while trying to debug this with him. Supposedly macOS uses "Apple Events" when launching an app through Finder, which calls the contained binary with its "application serial number"… or something? It supposedly does the same when using "Open with…" on a regular file & picking an app.

I know absolutely nothing about macOS. My initial response would be "we should talk to macOS to figure out what files BT was called on", but I don't know how much effort that will take without a local macOS machine to test everything on. A dirty workaround would be to ignore CLI arguments starting with `-psn_`: it'll prolly work for now but ideally it wouldn't be a long-term solution. I think this whole ordeal breaks loading modules at launch on macOS too?

## How to reproduce

1. (Re)move `~/.config/BambooTracker`
2. Launch BambooTracker.app via Finder

## System Information

- **Operating System**: Mac OS X 10.14.5
- **BambooTracker Version**: master b8cd8c88939710a143c9000d091f97b4656966c9
- **Build Type**: development build

rerrahkr:
close #282 

This PR move the position of `QApplication` instantiation to modify program arguments.

### DIFF
--- a/BambooTracker/gui/q_application_wrapper.cpp
+++ b/BambooTracker/gui/q_application_wrapper.cpp
@@ -27,7 +27,7 @@
 #include <exception>
 #include <QMessageBox>
 
-QApplicationWrapper::QApplicationWrapper(int& argc, char** argv) : QApplication (argc, argv) {}
+QApplicationWrapper::QApplicationWrapper(int& argc, char** argv) : QApplication(argc, argv) {}
 
 bool QApplicationWrapper::notify(QObject* receiver, QEvent* event)
 {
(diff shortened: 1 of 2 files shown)


N-SPC700 (approved):
Launches fine without any more module errors on first run.

rerrahkr:
Thanks for testing!

PR #288, opened 2020-11-15
Add linking options for RtAudio and RtMidi

### LINKED ISSUE #263: [Feature] - Option for linking against system Rt{Audio,Midi}
> #### Checklist
> 
> 
> 

---

## Requested Feature

An option to skip building the in-tree versions of RtAudio and RtMidi and link against their system-supplied dynamic libraries.

This could help with avoiding compilation failures caused by our pedantic error switches and is something that Linux distros would greatly prefer.

Provoked by this compilation failure in RtMidi someone sent me, specific to their 32-bit Linux CI:

```
midi/RtMidi/RtMidi.cpp: In function 'void* alsaMidiHandler(void*)':
midi/RtMidi/RtMidi.cpp:1612:25: error: comparison of integer expressions of different signedness: 'unsigned int' and '__syscall_slong_t' {aka 'long int'} [-Werror=sign-compare]
 1612 |           if (x.tv_nsec < y.tv_nsec) {
      |               ~~~~~~~~~~^~~~~~~~~~~
```

rerrahkr:
Close #263 

On executing `qmake`, you can use installed RtAudio or RtMidi to add `LIBS += -lrtaudio` or `LIBS += -lrtmidi` to the command line arguments (maybe)

I had another idea specifying it with `CONFIG` like selecting Audio APIs, but setting library links directly makes the change explicit.


OPNA2608:
> I had another idea specifying it with `CONFIG` like selecting Audio APIs, but setting library links directly makes the change explicit.

I think a `CONFIG` option would be conceptionally cleaner, other build systems I've used in the past sometimes have options like that (`--with-system-zlib`). Having to manually add a linker flag looks more like a workaround / oversight fix than something intentional.

BambooTracker/midi/midi.hpp
```diff
@@ -24,12 +24,13 @@
...
 
 #pragma once
-#include "RtMidi/RtMidi.hpp"
+
 #include <string>
 #include <vector>
 #include <memory>
 #include <mutex>
 #include <cstdint>
+#include "RtMidi.hpp"
```
OPNA2608, line 33: The original headers were simply `*.h`, we renamed our in-tree versions for the pedantic compiler settings. This fails to compile for me because it's looking for `RtMidi.hpp` instead of `RtMidi.h`.

BambooTracker/stream/audio_stream_rtaudio.cpp
```diff
@@ -26,7 +26,7 @@
 #include "audio_stream_rtaudio.hpp"
 #include <string>
 #include <vector>
-#include "RtAudio/RtAudio.hpp"
+#include "RtAudio.hpp"
```
OPNA2608, line 29: Same here, system-installed header is called "RtAudio.h".

OPNA2608 (changes requested):

OPNA2608:
I think we can set up the Ubuntu builder as a test for this linking option, I'll check which packages we'll need and try to update the CI script accordingly.

I believe this change will cause some problems with most RtMidi packages because we rely on RTMIDI_DUMMY as a safe fallback, so much so that we [modified our RtMidi.cpp to always define it](https://github.com/rerrahkr/BambooTracker/commit/7f1c31c82d0fb0aaa82bd0e478c3a5b9610b2e88#diff-5e2e927564f7a317b5c2ff872ce64d98b13d875ea181eb728cea733ef73e5e1e).

RtMidi system packages will most likely not have a patch/`-D` for it, so any `switchApi(RtMidi::RTMIDI_DUMMY)` will fail with "unsupported API". This needs to be accounted for so we don't reintroduce #172.

rerrahkr:
> I think a CONFIG option would be conceptionally cleaner

I try changing `"CONFIG += with_system_rtaudio with_system_rtmidi"` (or only `system_rtaudio system_rtmidi` is better?)

> I believe this change will cause some problems with most RtMidi packages because we rely on RTMIDI_DUMMY as a safe fallback, so much so that we [modified our RtMidi.cpp to always define it](https://github.com/rerrahkr/BambooTracker/commit/7f1c31c82d0fb0aaa82bd0e478c3a5b9610b2e88#diff-5e2e927564f7a317b5c2ff872ce64d98b13d875ea181eb728cea733ef73e5e1e).
> 
> RtMidi system packages will most likely not have a patch/`-D` for it, so any `switchApi(RtMidi::RTMIDI_DUMMY)` will fail with "unsupported API". This needs to be accounted for so we don't reintroduce #172.

so we need to modify the midi manager based on null-pointer check, not `RtMidi::RTMIDI_DUMMY` fallback. I will open another PR to tackle this

BambooTracker/BambooTracker.pro
```diff
@@ -551,8 +551,11 @@ INCLUDEPATH += \
     $$PWD/module \
     $$PWD/io
 
-include("stream/RtAudio/RtAudio.pri")
-include("midi/RtMidi/RtMidi.pri")
+# Linking settings
```
OPNA2608, line 554:
```suggestion
# Linking settings
system_*:CONFIG += link_pkgconfig
```

BambooTracker/BambooTracker.pro
```diff
@@ -551,8 +551,11 @@ INCLUDEPATH += \
...
+system_rtaudio:LIBS += -lrtaudio
```
OPNA2608, line 555:
```suggestion
system_rtaudio:PKGCONFIG += rtaudio
```

BambooTracker/BambooTracker.pro
```diff
@@ -551,8 +551,11 @@ INCLUDEPATH += \
...
+else:include("stream/RtAudio/RtAudio.pri")
 
-include("../data/data.pri")
+system_rtmidi: LIBS += -lrtmidi
```
OPNA2608, line 558:
```suggestion
system_rtmidi:PKGCONFIG += rtmidi
```

OPNA2608 (commented):
On my system I need to manually add rtmidi and rtaudio to `INCLUDEPATH` for the build to find the headers. We can use QMake's pkg-config integration to detect if a package exists, what flags to use for finding the headers & linking against the library and error out if the required package is missing. Unsure how well this works on other platforms.

`CONFIG+=system_rtmidi` without system-installed (pkg-config can't find it) rtmidi package:
```
Project MESSAGE: Qt configured with GCC
Project MESSAGE: Compiler is version 9.3.0
Project ERROR: rtmidi development package not found
```

rerrahkr:
Add pkgconfig settings, also rebased from master to fix an GitHub Actions error.

PR #289, opened 2020-11-15
Null-pointer based MIDI API check

rerrahkr:
#288 

replaced `RTMIDI_DUMMY` fallbacks

I fixed sources based on audio API initialization.

### DIFF
--- a/BambooTracker/configuration.cpp
+++ b/BambooTracker/configuration.cpp
@@ -183,6 +183,7 @@ Configuration::Configuration()
 	bufferLength_ = 40;
 
 	// Midi //
+	midiEnabled_ = false;
 	midiAPI_ = u8"";
 	midiInPort_ = u8"";
 
@@ -992,6 +993,10 @@ void Configuration::setBufferLength(size_t length) { bufferLength_ = length; }
 size_t Configuration::getBufferLength() const { return bufferLength_; }
 
 // Midi //
+void Configuration::setMidiEnabled(const bool enabled) { midiEnabled_ = enabled; }
+
+bool Configuration::getMidiEnabled() const { return midiEnabled_; }
+
 void Configuration::setMidiAPI(const std::string& api) { midiAPI_ = api; }
 
 std::string Configuration::getMidiAPI() const { return midiAPI_; }
--- a/BambooTracker/configuration.hpp
+++ b/BambooTracker/configuration.hpp
@@ -277,11 +277,14 @@ private:
 
 	// Midi //
 public:
+	void setMidiEnabled(const bool enabled);
+	bool getMidiEnabled() const;
 	void setMidiAPI(const std::string& api);
 	std::string getMidiAPI() const;
 	void setMidiInputPort(const std::string& port);
 	std::string getMidiInputPort() const;
 private:
+	bool midiEnabled_;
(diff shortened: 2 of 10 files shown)


OPNA2608:
Runs on a system without any exposed MIDI API, but brings up an "Unable to initialize MIDI!" error box on launch & any configuration change (similar to https://github.com/rerrahkr/BambooTracker/issues/172#issuecomment-569419413). There also is no option in the settings to just turn off MIDI (like with the Dummy API option previously) and thus no way of disabling this.

Would it be possible to either have a virtual "Off" API or (in my opinion cleaner) a checkbox to disable MIDI input (including the MIDI input options)?

rerrahkr:
I think it's easy to implement the checkbox. I also try to change it to force midi input to be disabled if BT get a midi error.

OPNA2608:
When I manually tick the MIDI checkbox, set an API that fails to initialize and press [OK] (doesn't happen with [Apply]?), the option disables itself after closing the error message. Is this intentional or accidentally being treated like a first launch & falling back to the "off" default? It feels inconsistent with how every other configuration option works. Audio in particular keeps the selected not-working API option instead which I think is preferable.

rerrahkr:
BT intentionally turned off it when an error occurred, but i will change it. MIDI input is set to disabled on first launch, so it probably doesn't need the MIDI API fallback, and leave the internal value of config unchanged even if the MIDI API is in error.

OPNA2608:
One last thing is that the device combobox doesn't get properly disabled when opening the configuration window. Toggling the MIDI checkbox to [x] -> [ ] once properly disables it.
grafik

> MIDI input is set to disabled on first launch

If possible, it'd be nice if the MIDI APIs could get automatically tested for a functional one on first launch, with "disabled" being the default if none of them are determined to be working. Similar to what we're doing with RtAudio but for RtMidi instead, so everything "Just Works:tm:" for new users and existing users transitioning to the next release won't be stumped by their MIDI device suddenly not working. We can move that into another feature request though if you'd rather just get the base functionality working for now, LGTM aside from this.

rerrahkr:
Audio streaming is a must-have feature for everyone, but MIDI input is only used by people who have a MIDI keyboard, so I think it is better to turn off at the first time. Many software I know also set up MIDI I/O manually after startup.

OPNA2608:
Hm, okay.

b208d50 didn't fix the problem afaict, combobox is still enabled upon opening Configuration dialogue. Qt5.9.5 & Linux if that helps with identifying the problem.

rerrahkr:
> combobox is still enabled upon opening Configuration dialogue.

I was confused because such a problem did not occur on Windows, but I was able to reproduce it on macOS, so I fixed it. I'm not sure why it happens, maybe it's a Qt bug.

OPNA2608:
Works but prints this on the terminal:

```
QMetaObject::connectSlotsByName: No matching signal for on_midiApiComboBox_currentIndexChanged(QString,bool)
```

rerrahkr:
Oh I didn't notice the log. I've separated its method so I think this is okay.

OPNA2608 (approved):

rerrahkr:
Thanks reviewing!

PR #290, opened 2020-11-17
Revert "Fix .vgi import to load the operator 2 and 3 correctly"

OPNA2608:
This reverts commit 8e7879fd9dcc66a16135c639747dea7a7209ccb7. The "incorrect" operator order is actually correct. Chip-internal order of values is 0->2->1->3, VGI stores the values in the order 0->1->2->3 just like TFI (which was unaffected by the reverted commit). See [VGI File Format](https://vgmrips.net/wiki/VGI_File_Format) and [TFI File Format](https://vgmrips.net/wiki/TFI_File_Format).

### DIFF
--- a/BambooTracker/io/instrument_io.cpp
+++ b/BambooTracker/io/instrument_io.cpp
@@ -2908,20 +2908,6 @@ AbstractInstrument* InstrumentIO::loadVGIFile(const BinaryContainer& ctr, std::s
 	ssgeg1 = ssgeg1 & 8 ? ssgeg1 & 7 : -1;
 	instManLocked->setEnvelopeFMParameter(envIdx, FMEnvelopeParameter::SSGEG1, ssgeg1);
 
-	instManLocked->setEnvelopeFMParameter(envIdx, FMEnvelopeParameter::ML2, ctr.readUint8(csr++));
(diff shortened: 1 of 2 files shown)


rerrahkr:
I checked TFM Music Maker about tfi, and yes, BT has wrong order so I'll fix it.

With regard to vgi, It's strange. I made vgi by VGM Music Maker (v1.1) and fixed BT to load it correctly at 8e7879f.
[test.zip](https://github.com/rerrahkr/BambooTracker/files/5560364/test.zip) - I made an example vgi and load it on VGM Music Maker and BambooTracker:

load vgi

OPNA2608:
I understand, then the VGI files that were tried must have the wrong order. I'll close this, maybe a simple tool to fix incorrect VGI/TFI files en masse would be useful to some users :thinking:.

OPNA2608:
> load vgi

Please note that the algorithm graphs between VGM Music Maker and BambooTracker also exhibit this wrong order: 2 is 3, 3 is 2. Thus your fix most definitely was incorrect. and needs to be reverted.

VGM MM example:
grafik

Expected & correct operator data if imported into BT: (sounds the same)
grafik

Actual operator data if imported into BT: (sounds different than in VGM MM)
grafik

PR #291, opened 2020-11-23
Fix icon install paths

OPNA2608:
Should be `hicolor/{res}/apps/` instead of `hicolor/{res}/` as mentioned on the Discord server. See a user's patch on the AUR build script [here](https://aur.archlinux.org/packages/bambootracker-git/#comment-774462) which fixes this downstream. (among other things which shouldn't be required anymore, but oh well ¯\\\_(ツ)\_/¯)


rerrahkr:
Thanks for your contribution!

PR #294, opened 2020-12-09
Use submodules and subdirs

OPNA2608:
(targeting link_rtaudio_rtmidi branch for lower code delta)

- Rt{Audio,Midi} have been deleted from the repository & re-added as git submodules.
- A root-level QMake project file "Project.pro" (good enough name-wise, though open for suggestions) has been added. It uses the `subdirs` template to check & build the sub-projects:
  - BambooTracker (C++ code for the tracker itself)
  - data (assets that aren't compiled into object files)
  - RtAudio (handles feature detection for & statically compiling of RtAudio, disabled by passing `CONFIG+=system_rtaudio`)
  - RtMidi (handles feature detection for & statically compiling of RtMidi, disabled by passing `CONFIG+=system_rtaudio`)
- The resource pack data (`/data/resources/`) has been moved to `/BambooTracker`. While it's technically better in `/data`, it's being compiled into an \*.o file which doesn't fit well with how the rest of the assets are being treated (no compiling & linking into the BambooTracker binary required, so using compiler-less `TEMPLATE` option) and makes including the object file into BambooTracker abit harder than when everything was mashed together. Open for suggestions here.
- Added automatic feature detection to the Rt{Audio,Midi} sub-projects. A series of `qtCompileTest`s is used to determine what APIs are supported by the system, similar to how Rt{Audio,Midi}'s own build system detects available features.

~~needs to be retargeted at master after #288 is merged~~ See message below.


OPNA2608:
Updated testing to take the platform's required & specifically enabled APIs into consideration, erroring out early of either of them cannot be satisfied. Needs more changes to disable all this testing if `system_rt*` is passed.

OPNA2608:
@rerrahkr I consider this basically-done now, I'd appreciate if you could try to compile & run this on Windows & possibly macOS. I've struggled abit with the added git submodules at the start, I hope they're not giving you any problems (might need a `git submodule init && git submodule update` after checking out for `submodules/Rt{Audio,Midi}/src` to get the repo contents).

I don't know which branch to merge this into - it's based on `link_rtaudio_rtmidi` but doing it like this won't run any build tests, merging it in `master` would be better for the added github & appveyor builds but there are merge conflicts I'll have to look into first, some of which may be inherited from your branch.

For the time being, I'll push this branch to my fork's master and push a tag there, bearing in mind that the XP builds will fail due to #296.

OPNA2608:
GitHub CI: https://github.com/OPNA2608/BambooTracker/runs/1581887247
AppVeyor CI: https://ci.appveyor.com/project/OPNA2608/bambootracker/builds/36921263
Release: https://github.com/OPNA2608/BambooTracker/releases/tag/v0.0.0

macOS on GitHub failed, looks like a problem with Homebrew (which I didn't really touch since the last run). Unfortunate, I hope that fixes itself without me having to touch the Homebrew stuff. :confused:

rerrahkr:
I will test them on mac later.

When I run the Windows build of Appveyor, I get an error that Qt5Core.dll and Qt5Gui.dll are missing.
It seems that there are two Qt dlls, one for release (Qt5Core.dll) and one for debugging (Qt5Cored.dll) ([found it here](https://stackoverflow.com/a/18161647)), so the release build setting may be enabled.

.appveyor.yml
```diff
@@ -71,26 +71,27 @@ platform: x86
...
       export JACKLIB="/usr/local/opt/jack/lib"
       export JACKINC="/usr/local/opt/jack/include"
-  - cd BambooTracker
+      export PKG_CONFIG_PATH="/usr/local/opt/jack/lib/pkgconfig"${PKG_CONFIG_PATH:+':'}$PKG_CONFIG_PATH
 
 # to run your custom scripts instead of automatic MSBuild
 build_script:
   - ps: |
-      qmake.exe BambooTracker.pro PREFIX=..\target CONFIG+=debug CONFIG+=console CONFIG+=nostrip
+      qmake.exe Project.pro PREFIX=..\target CONFIG+=debug CONFIG+=console CONFIG+=nostrip
```
rerrahkr, line 83: Should we add `CONFIG-=release` to explicitly indicate that it is a debug?

OPNA2608: Just to be safe I added, that but it was most likely `PREFIX=..\target` now being wrong since we now have a top-level project file instead of `cd`ing into the Bambootracker directory

OPNA2608:
Hmm nevermind, that didn't fix it… I'll dig out my Windows machine and try to figure out where those dependencies on release libraries are coming from. :confused: 

Before:
> [00:06:30] readPeExecutable: C:\projects\bambootracker\target\BambooTracker.exe 32 bit, MinGW, dependent libraries: 12, debug
[00:06:30] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Cored.dll 32 bit, MinGW, dependent libraries: 11
[00:06:30] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Guid.dll 32 bit, MinGW, dependent libraries: 7
[00:06:30] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Widgetsd.dll 32 bit, MinGW, dependent libraries: 9

After:
> [00:06:27] readPeExecutable: C:\projects\bambootracker\target\BambooTracker.exe 32 bit, MinGW, dependent libraries: 12, debug
[00:06:27] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Core.dll 32 bit, MinGW, dependent libraries: 11
[00:06:27] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Gui.dll 32 bit, MinGW, dependent libraries: 7
[00:06:27] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Widgetsd.dll 32 bit, MinGW, dependent libraries: 9
[00:06:27] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Cored.dll 32 bit, MinGW, dependent libraries: 11
[00:06:27] readPeExecutable: C:\Qt\5.5\mingw492_32\bin\Qt5Guid.dll 32 bit, MinGW, dependent libraries: 7

rerrahkr:
> I will test them on mac later.

I tested them a while ago, and they worked fine.

OPNA2608:
GitHub CI: https://github.com/OPNA2608/BambooTracker/commit/e2d189741e8986e6ebc1c8031645ab55350b6745/checks
AppVeyor CI: https://ci.appveyor.com/project/OPNA2608/bambootracker/builds/36982650
Same release. Qt5 library linking and macOS on GHA should be fixed. It's perhaps not very pretty, but it seems to work.

.github/workflows/macos.yml
```diff
@@ -48,30 +48,39 @@ jobs:
...
       - name: Installing dependencies.
         run: |
-          brew install qt jack p7zip
-          brew link --force qt
+          brew update
+          # JACK implies brew link python-3.9, fails due to shipped python binaries
+          brew unlink python@3.9
+          brew link --overwrite python@3.9
+          brew upgrade --force python@3.9
+          brew install qt5 pkg-config jack p7zip
```
OPNA2608, line 66: `jack` has a dependency on `python@3.9`. When it tries to install `python-3.9.1_1` on GHA and link its binaries into the system's common bin path, it finds a symlink/binary(?) in /usr/local/bin that belongs to macOS' shipped python2 installation and errors out. The suggested options to instead overwrite it assume that there was an attempt to install just `python@3.9` and I couldn't get it to work for this dependency use-case.

```
==> Pouring python@3.9-3.9.1_1.catalina.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3
Target /usr/local/bin/2to3
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.9

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.9

Possible conflicting files are:
/usr/local/bin/2to3 -> /Library/Frameworks/Python.framework/Versions/2.7/bin/2to3
```

Unlinking the pre-installed `python@3.9` package and relinking it somehow, magically, fixes this. The `brew upgrade` is just to make sure that it's not python that's failing anymore.

submodules/common/library.pri
```diff
@@ -0,0 +1,44 @@
+TEMPLATE = lib
+CONFIG -= debug_and_release debug_and_release_target
```
OPNA2608, line 2: Windows builds seem to have this enabled by default. Along with Shadow Builds, which are also enabled by default on Qt5 Windows, the build process will create `/release` and `/debug` directories for any build. These settings will additionally permeate into the generated `*.prl` files, which accompany the compiled libraries and contain importable compilation settings for the BambooTracker subproject. I use these to inherit the detected and explicitly required library dependencies and their required linking flags.

During prl creation, linking flags for Qt5's release libraries along with these two CONFIG settings find their way into the linker settings, even if `CONFIG-=release CONFIG+=debug` is specified. Unsettings these this way fixes the problem. We could use some more sophisticated Qmake logic to instead properly support shadow builds into, reading the correct prl file from and linking against the right vendored Rt\* library from the appropriate `/submodules/Rt{Audio,Midi}/{release,debug}` directory? The current way, if I'm reading the docs correctly, may cause problems when switching between release and debug configurations without `make clean`ing the affected subprojects… But I'm fine with this for now. :slightly_smiling_face:

rerrahkr: In submodules/common/library.pri:

```qmake
DESTDIR = $$_PRO_FILE_PWD_
OBJECTS_DIR = $$_PRO_FILE_PWD_
MOC_DIR = $$_PRO_FILE_PWD_
```

If TARGET is the same for debug and release, it seems that these settings make qmake generate a prl including both debug and release settings.

While taking advantage of the shadow build, I added "d" suffix to the library names in the case of debug so that the binaries would not be mixed.
RtMidi:
- Dubug: librtmidid.a
- Release: librtmidi.a

[patch.txt](https://github.com/rerrahkr/BambooTracker/files/5742877/patch.txt)

OPNA2608: Perfect, committed in 34835df!

rerrahkr:
LGTM. Would you mind me merging this PR?
After merging, I will try to resolve the conflict with `master` in #288.

OPNA2608:
Go ahead. :+1:

rerrahkr:
Thanks for your contribution!

PR #307, opened 2021-01-11
Fix macOS GHA

OPNA2608:
GitHub CI complains about p7zip already being installed but outdated. We don't depend on the latest p7zip so the pre-installed one will be fine.


rerrahkr:
Thanks for your contribution!

PR #324, opened 2021-02-08
Fix NOTE_NONE getting transposed

EnfauKerus:
Was encountering transposing problem and narrowed it down here.  
After checking `hasGeneralNote()` usage between old revision, I think this is what it should be.  
point me out if I got something wrong :ok_hand:


rerrahkr:
It works well. Thanks for your contribution!

PR #332, opened 2021-02-17
Qt6 support & building

### LINKED ISSUE #298: [Feature] - Qt6 support
> #### Checklist
> 
> 
> 

---

## Requested Feature

Qt6 was publicly released just afew days ago! I've already tried to build BT with the win64_mingw81 version of Qt6 and had decent success after some minimal code changes:

- gui/configuration_dialog.cpp:213:
`void QLayout::setMargin(int margin)` was removed, `void QLayout::setContentsMargins(int left, int top, int right, int bottom)`works
- gui/mainwindow.cpp:43:
`QDesktopWidget` was removed, I was able to remove the offending include statement without any problems
- gui/order_list_editor/order_list_panel.cpp:108 & gui/pattern_editor/pattern_editor_panel.cpp:157:
`QFont::ForceIntegerMetrics` was removed, I didn't find a 1:1 replacement for it so I used `QFont::PreferMatch` instead
- gui/mainwindow.cpp:48:
`QTextCodec` was replaced in direct favour of `QStringConverter`, but continues to exist in the new Qt5Compat module for porting ease. [See this blog entry for details](https://www.qt.io/blog/porting-from-qt-5-to-qt-6-using-qt5compat-library). I added Qt5Compat to the build for porting ease though defines could be used to mitigate the need for it. This would be preferable because I had to add some extra handling to the qt fetching script to handle Qt5Compat's weird location in the sdk repo.
- I also added some extra flags to not fail on every new deprecation warning, though those should ideally be fixed. Maybe via Qt6-specific defines too?

I have a branch with those changes [here](https://github.com/OPNA2608/BambooTracker/tree/gha_qt6). There are some minor problems left:

- The WASAPI capability detection in RtAudio.pri tests for `greaterThan(QT_MAJOR_VERSION, 4):greaterThan(QT_MINOR_VERSION, 5)` which doesn't handle `6` and `0` gracefully, resulting in a DirectSound-only build.
- For some reason windeployqt missed the MinGW DLLs resulting in a broken artifact. Maybe this is because I used the system-installed compiler instead of also rewriting the mingw-fetcher to have support for 64-bit versions? I'm unsure.
- For convenience with debugging & wait time I also dropped all CI except modern Windows, that would need to be readded / improved for Qt6 as well.

I wouldn't consider this idea of significant importance yet, but eventually Qt5 will become deprecated & unsupported just like Qt4 before it, so it can't hurt to keep compatibility in mind.

OPNA2608:
Closes #298.

The goal here is to get BambooTracker compiled & running on Qt6. I don't think moving the regular Windows CI to Qt6 is a good idea because it only supports 64-bit Windows 10, so I made a new GHA target.

("Windows 7 and up" should eventually be downgraded to 5.12, I noticed. 5.15 is only a LTS release for commercial Qt users and was only supported until Qt6 was released. 5.12 is still under free long-term support until December 2021)

Compared to #298, I've bumped the Qt6 version to yesterday's Qt6.1.0 Alpha release and fixed the MinGW script to support 64-bit and 32-bit fetching.

TODOs:

- [x] Get it compiled on CI with minimal changes
- [x] Check if `windeployqt` finds all MinGW DLLs, figure out why not if it doesn't
- [x] Fix WASAPI capability detection in RtAudio.pro
- [x] Test on a supported system

Future TODOs:
- [ ] Build without Qt5 compatibility module (`qt5compat`)
- [x] Build without warning disable flags
  - [x] `-Wno-error=deprecated-declarations`
  - [x] `-Wno-error=unused-result`
- [x] Check Darwin Qt6 once GHA/Homebrew forces it on us

### DIFF
--- a/.github/workflows/windows-10-qt6.yml
+++ b/.github/workflows/windows-10-qt6.yml
@@ -22,7 +22,7 @@ defaults:
     shell: bash
 
 env:
-  BUILD_TARGET: windows-10-x86_64-qt6
+  BUILD_TARGET: windows-10-x86_64
   MAKE: mingw32-make
   BT_INSTALLBASE: ${{ github.workspace }}/target/
   MINGW_VERSION: 8.1.0
--- a/.github/workflows/windows-10-qt6.yml
+++ b/.github/workflows/windows-10-qt6.yml
@@ -22,7 +22,7 @@ defaults:
     shell: bash
 
 env:
-  BUILD_TARGET: windows-10-x86_64
+  BUILD_TARGET: windows-10-64bit
   MAKE: mingw32-make
   BT_INSTALLBASE: ${{ github.workspace }}/target/
   MINGW_VERSION: 8.1.0
--- a/.github/workflows/windows-10-qt6.yml
+++ b/.github/workflows/windows-10-qt6.yml
@@ -29,7 +29,7 @@ env:
   MINGW_BITNESS: 64
   MINGW_CHOCOBASE: /c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64
   MINGW_INSTALLBASE: ${{ github.workspace }}/mingw64/
-  QT_VERSION: 6.1.0
+  QT_VERSION: 6.0.2
   QT_TOOLCHAIN: win64_mingw81
   QT_MODULES: qtbase qttools qttranslations qt5compat
   QT_INSTALLBASE: ${{ github.workspace }}/Qt/
--- a/.github/workflows/windows-10-qt6.yml
+++ b/.github/workflows/windows-10-qt6.yml
@@ -1,4 +1,4 @@
-name: Windows (7 and up, 32-bit, Qt5)
+name: Windows 10 (64-bit, Qt6)
 
 on:
   push:
@@ -22,16 +22,16 @@ defaults:
     shell: bash
 
 env:
-  BUILD_TARGET: windows
+  BUILD_TARGET: windows-10-x86_64-qt6
   MAKE: mingw32-make
   BT_INSTALLBASE: ${{ github.workspace }}/target/
   MINGW_VERSION: 8.1.0
-  MINGW_BITNESS: 32
-  MINGW_CHOCOBASE: /c/ProgramData/chocolatey/lib/mingw/tools/install/mingw32
-  MINGW_INSTALLBASE: ${{ github.workspace }}/mingw32/
-  QT_VERSION: 5.15.1
-  QT_TOOLCHAIN: win32_mingw81
-  QT_MODULES: qtbase qttools qttranslations
+  MINGW_BITNESS: 64
+  MINGW_CHOCOBASE: /c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64
+  MINGW_INSTALLBASE: ${{ github.workspace }}/mingw64/
+  QT_VERSION: 6.1.0
+  QT_TOOLCHAIN: win64_mingw81
+  QT_MODULES: qtbase qttools qttranslations qt5compat
   QT_INSTALLBASE: ${{ github.workspace }}/Qt/
 
 jobs:
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -7,6 +7,8 @@
 QT       += core gui
 
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
+# Porting helper Qt5 -> Qt6
+greaterThan(QT_MAJOR_VERSION, 5): QT += core5compat
 
 TARGET = BambooTracker
 TEMPLATE = app
@@ -78,6 +80,11 @@ else:clang|if(gcc:!intel_icc) {
   message("Compiler is version" $$COMPILER_VERSION)
 
   # Temporary known-error downgrades here
+
+  # WIP Requires a deeper look into the warnings to make Qt5 -> Qt6 transition clean
+  greaterThan(QT_MAJOR_VERSION, 5) {
+    CPP_WARNING_FLAGS += -Wno-error=deprecated-declarations -Wno-error=unused-result
+  }
 }
 else {
   message("Configured compiler is unknown, no attempt to add warning & pedantic compiler switches")
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -80,11 +80,6 @@ else:clang|if(gcc:!intel_icc) {
   message("Compiler is version" $$COMPILER_VERSION)
 
   # Temporary known-error downgrades here
-
-  # WIP Requires a deeper look into the warnings to make Qt5 -> Qt6 transition clean
-  greaterThan(QT_MAJOR_VERSION, 5) {
-    CPP_WARNING_FLAGS += -Wno-error=deprecated-declarations -Wno-error=unused-result
-  }
 }
 else {
   message("Configured compiler is unknown, no attempt to add warning & pedantic compiler switches")
--- a/BambooTracker/gui/configuration_dialog.cpp
+++ b/BambooTracker/gui/configuration_dialog.cpp
@@ -223,7 +223,7 @@ ConfigurationDialog::ConfigurationDialog(std::weak_ptr<Configuration> config, st
 		QObject::connect(button, &QToolButton::clicked, seq, &QKeySequenceEdit::clear);
 		auto layout = widget->layout();
 		layout->setSpacing(0);
-		layout->setMargin(0);
+		layout->setContentsMargins(0, 0, 0, 0);
 		layout->addWidget(seq);
 		layout->addWidget(button);
 		ui->shortcutsTreeWidget->setItemWidget(item, 1, widget);
--- a/BambooTracker/gui/instrument_editor/visualized_instrument_macro_editor.cpp
+++ b/BambooTracker/gui/instrument_editor/visualized_instrument_macro_editor.cpp
@@ -406,7 +406,7 @@ void VisualizedInstrumentMacroEditor::drawLoop()
 					painter.fillRect(w, loopY_, 2, fontHeight_, palette_->instSeqLoopEdgeColor);
 					QString times = (loops_[j].times == 1) ? "" : QString::number(loops_[j].times);
 					painter.drawText(QRectF(w + 2, loopY_, colWidth - 2, fontHeight_),
-									 Qt::AlignLeft + Qt::AlignVCenter, tr("Loop %1").arg(times));
+									 Qt::AlignLeft | Qt::AlignVCenter, tr("Loop %1").arg(times));
 				}
 				if (loops_[j].end == i) {
 					painter.fillRect(w + colWidth - 2, loopY_, 2,
--- a/BambooTracker/gui/instrument_editor/visualized_instrument_macro_editor.cpp
+++ b/BambooTracker/gui/instrument_editor/visualized_instrument_macro_editor.cpp
@@ -33,7 +33,7 @@
 #include <QApplication>
 #include <QFontMetrics>
 #include <QPainter>
-#include <QPoint>
+#include <QPointF>
 #include <QRegularExpression>
 #include <QRegularExpressionMatch>
 #include "gui/event_guard.hpp"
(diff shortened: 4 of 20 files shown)


OPNA2608:
Almost done, just needs a tiny bit more cleaning up I think.

> - [ ] Build without Qt5 compatibility module (`qt5compat`)

I looked into this and it seems basically-unviable. BT uses `QTextCodec` to translate to & from Shift-JIS encoding. Qt6 deprecated `QTextCodec` by moving it into the `qt5compat` module, the replacement class `QStringConverter` supports alot fewer codecs though.

Compare:
grafik

I'm not sure what the intended way of reading/writing most non-UTF codecs under Qt6 is supposed to be if `QTextCodec` is considered a remnant of Qt5. I guess you just write your own en-/decoder :confused:?

BambooTracker/gui/configuration_dialog.cpp
```diff
@@ -223,7 +223,7 @@ ConfigurationDialog::ConfigurationDialog(std::weak_ptr<Configuration> config, st
 		QObject::connect(button, &QToolButton::clicked, seq, &QKeySequenceEdit::clear);
 		auto layout = widget->layout();
 		layout->setSpacing(0);
-		layout->setMargin(0);
+		layout->setContentsMargins(0, 0, 0, 0);
```
OPNA2608, line 226: `QLayout::setMargin(int)` is obsoleted in Qt5 & was removed in Qt6, `QLayout::setContentsMargins(int, int, int, int)` or `QLayout::setContentsMargins(QMargins)` are the simple replacement options.

BambooTracker/gui/instrument_editor/visualized_instrument_macro_editor.cpp
```diff
@@ -406,7 +406,7 @@ void VisualizedInstrumentMacroEditor::drawLoop()
 					painter.fillRect(w, loopY_, 2, fontHeight_, palette_->instSeqLoopEdgeColor);
 					QString times = (loops_[j].times == 1) ? "" : QString::number(loops_[j].times);
 					painter.drawText(QRectF(w + 2, loopY_, colWidth - 2, fontHeight_),
-									 Qt::AlignLeft + Qt::AlignVCenter, tr("Loop %1").arg(times));
+									 Qt::AlignLeft | Qt::AlignVCenter, tr("Loop %1").arg(times));
```
OPNA2608, line 409: Qt flags should be combined via bitwise OR operations, adding them throws a compile-time exception with Qt6.

BambooTracker/gui/mainwindow.cpp
```diff
@@ -41,7 +41,6 @@
 #include <QMimeData>
 #include <QProgressDialog>
 #include <QRect>
-#include <QDesktopWidget>
```
OPNA2608, line 44: `QDesktopWidget` is obsolete in Qt5 & was removed in Qt6. I didn't see anything that actually used this include so I just removed it & everything appears to still work fine. If this was intended for some incomplete code, refer to Qt6's docs: https://doc.qt.io/qt-6/widgets-changes-qt6.html#qdesktopwidget-and-qapplication-desktop

BambooTracker/gui/order_list_editor/order_list_panel.cpp
```diff
@@ -105,7 +105,7 @@ OrderListPanel::OrderListPanel(QWidget *parent)
 	headerFont_.setPointSize(10);
 	rowFont_ = QFont("Monospace", 10);
 	rowFont_.setStyleHint(QFont::TypeWriter);
-	rowFont_.setStyleStrategy(QFont::ForceIntegerMetrics);
+	rowFont_.setStyleStrategy(QFont::PreferMatch);
```
OPNA2608, line 108: `QFont::ForceIntegerMetrics` was removed in Qt6. I couldn't find a perfect 1:1 replacement from the docs so I picked `QFont::PreferMatch`. If the choice of `QFont::ForceIntegerMetrics` was a deliberate one and `QFont::PreferMatch` breaks something visually, please let me know.

BambooTracker/gui/pattern_editor/pattern_editor_panel.cpp
```diff
@@ -157,7 +157,7 @@ PatternEditorPanel::PatternEditorPanel(QWidget *parent)
 	headerFont_.setPointSize(10);
 	stepFont_ = QFont("Monospace", 10);
 	stepFont_.setStyleHint(QFont::TypeWriter);
-	stepFont_.setStyleStrategy(QFont::ForceIntegerMetrics);
+	stepFont_.setStyleStrategy(QFont::PreferMatch);
```
OPNA2608, line 160: Same as in `BambooTracker/gui/order_list_editor/order_list_panel.cpp`.

BambooTracker/BambooTracker.pro
```diff
@@ -78,6 +80,11 @@ else:clang|if(gcc:!intel_icc) {
   message("Compiler is version" $$COMPILER_VERSION)
 
   # Temporary known-error downgrades here
+
+  # WIP Requires a deeper look into the warnings to make Qt5 -> Qt6 transition clean
+  greaterThan(QT_MAJOR_VERSION, 5) {
+    CPP_WARNING_FLAGS += -Wno-error=deprecated-declarations -Wno-error=unused-result
+  }
```
OPNA2608, line 87: I hope we can get rid of these again in the future, possibly without any fancy `#ifdef QT6 … #else … #endif` stuff. :slightly_smiling_face:

.github/workflows/windows-10-qt6.yml
```diff
@@ -0,0 +1,156 @@
...
+      - '**.txt'
+      - 'LICENSE'
+      - '.gitignore'
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  BUILD_TARGET: windows-10-x86_64-qt6
```
OPNA2608, line 25: Is this too unwieldy of a release package name? I was hoping to avoid any confusion about what zip to download by making this as specific as possible.

OPNA2608:
I think I still need to adjust the README to not be Qt5-specific.

OPNA2608:
Updated the READMEs, I hope the changes to README_jp are correct.

OPNA2608:
Summary of the leftover warnings:

- methods of Q\*Event classes that were used to get position parameters (`x()` & `pos()`) should use `position()` in Qt6, sometimes `position().toPoint()`?
- `QApplication::setAttribute(Qt::AA_EnableHighDpiScaling)` is deprecated in Qt6, High-DPI scaling is enabled by default there
- `QTranslator::load`'s return value is marked as non-discardable now

I think I'm done with this for now though, feel free to review it now. This'll add a new 64-bit, Windows 10 build test & release target with Qt6. I don't know how that may be incorporated into the project page's platform-detection, if at all? It should also be tested on a Linux system with Qt6 packages, though I don't have one of those installed right now.

rerrahkr:
[patch.zip](https://github.com/rerrahkr/BambooTracker/files/6091514/patch.zip)
I used the return value of `QTranslator::load()` as a flag for the debug output. Also, I changed `pos()`s accordingly.
As for High-DPI, before Qt5, you can't delete that line because it doesn't work without setting the attribute. 
Since we are using `QT_VERSION_CHECK`:
```c++
#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
		QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
#endif
```
so I think we can add whether Qt6 or not to the condition.

---

By the way, I don't know how to set the High-DPI setting in my environment, so I doubt if this is working well in Qt5.

rerrahkr:
Appveyor warned using deprecated method:
```
main.cpp:128:23: error: 'location' is deprecated: Use path() [-Werror,-Wdeprecated-declarations]
        return QLibraryInfo::location(QLibraryInfo::TranslationsPath);
                             ^
```

It will work if you add a if statement:
```c++
#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
	return QLibraryInfo::location(QLibraryInfo::TranslationsPath);
#else
	return QLibraryInfo::path(QLibraryInfo::TranslationsPath);
#endif
```

OPNA2608:
Thanks for the hint, I didn't check AppVeyor! I don't remember if that's intentional or an oversight from afew months ago: we have Homebrew's qt pinned to 5 on GHA but unpinned (and hence -> 6) on AppVeyor. Should we keep it like this or use the same Qt version for development builds and releases? (Qt6 bumps the minimum macOS version from High Sierra to Mojave)

I'm clueless about the sudden build failure on Windows Qt6, it packaged the Qt translations fine on d6664e2 but started failing on 3b78931… The log order is abit messed up but I assume it's this:

```
Creating qt_ar.qm...
Running: lconvert -o D:\a\BambooTracker\BambooTracker\target\translations\qt_ar.qm qtbase_ar.qm
QM-Format error

Error: Process completed with exit code 1.
```

OPNA2608:
Go figure, bumping the Qt6 version down to the latest stable one fixed that. :sweat_smile:

OPNA2608:
- [ ] What Qt version (5.x or 6.x) should we use for macOS releases / development builds?
- [ ] What should the Qt6 64-bit Windows build be called in the release section? Maybe just `windows-10-x86_64`?
- [ ] Add explanation about all this "'Windows' vs 'Windows 10'" release stuff to READMEs
- [ ] Implement the new release in the website's download script?

rerrahkr:
> * What Qt version (5.x or 6.x) should we use for macOS releases / development builds?

I won't be able to borrow my macbook from my friends anymore, so I can't test it locally without buying it. I think it is better to keep Qt5 to avoid encountering unexpected errors.

> * What should the Qt6 64-bit Windows build be called in the release section? Maybe just `windows-10-x86_64`?

That's fine for me.

> * Add explanation about all this "'Windows' vs 'Windows 10'" release stuff to READMEs

It is not necessary I think. Changing the job names of Travis CI would make the difference obvious like those of GitHub Actions. it would also be better to specify that the traditional "Windows" is a 32-bit version.

> * Implement the new release in the website's download script?

We need to change it because the webpage queries GitHub API for the release binary names. This should also display "xx-bit" in items of the download button.

OPNA2608:
> I won't be able to borrow my macbook from my friends anymore, so I can't test it locally without buying it. I think it is better to keep Qt5 to avoid encountering unexpected errors.

I have a macOS test environment now & can run Bambootracker successfully, but it's only High Sierra. I'll pin the macOS development builds to Qt5 then. I don't know when Homebrew will drop Qt5 completely but I'll look into upgrading to whatever's still supported by Qt6 when that happens.

> That's fine for me.

Okay, will do.

> It is not necessary I think. Changing the job names of Travis CI would make the difference obvious like those of GitHub Actions. it would also be better to specify that the traditional "Windows" is a 32-bit version.

I don't quite understand, sorry. We have dropped Travis Ci a while ago. Would you like me to adjust the traditional "Windows" release's name to ~ `windows-x86` to complement the `windows-10-x86_64` one? If that's perhaps too technical/confusing for a regular user, maybe `windows-7-32bit` and `windows-10-64bit`?

> We need to change it because the webpage queries GitHub API for the release binary names. This should also display "xx-bit" in items of the download button.

I'll look into it.

OPNA2608:
I hope everything's done? I've pinned AppVeyor macOS to `qt5`, added an architecture suffix to every kind of artefact (`-x86` / `-x86_64`) and renamed the two CI jobs to `Windows 7 and up (32-bit, Qt5)` and `Windows 10 (64-bit, Qt6)`.

When this is merged, I'll work on another PR against the `gh-pages` branch to fix the links when they have the architecture prefix & add platform detection/download link for Windows 10 64-bit.

rerrahkr:
Sorry for my late reply.

> Would you like me to adjust the traditional "Windows" release's name to ~ `windows-x86` to complement the `windows-10-x86_64` one? If that's perhaps too technical/confusing for a regular user, maybe `windows-7-32bit` and `windows-10-64bit`?

I think the later ones may be easier for non-coders to understand. I think they rarely get the dev builds, so we only need to modify release builds names.

OPNA2608:
No problem!

I understand, I'll adjust the naming. I don't see any harm in changing the dev builds' names like this. "64bit" might still be confusing for macOS desktop users with Apple's new 64-bit ARM CPUs, but I think we can worry about that another time - I wouldn't want to call it "macos-intel" or something.

rerrahkr:
There seems to be nothing wrong with the job itself, just not enough storage space for the artifacts. Would you mind me merging this PR?

OPNA2608:
Please go ahead if you think it's fine. :rocket:

rerrahkr:
Thanks for your contribution!

PR #334, opened 2021-02-17
CI improvements

OPNA2608:
Requires https://github.com/NixOS/nixpkgs/pull/113415.

I've updated the build expression on Nixpkgs for 0.4.6 accordingly, just need the update to get merged into stable. Until then you can disregard this PR, I'll mark it as ready eventually.

Alternatively we could switch to the unstable channel and merge this now but I would like to avoid doing future binary releases on a unstable channel or switching between stable & unstable all the time.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -95,7 +95,9 @@ platform: x86
 before_build:
   - ps: $env:Path = "$env:QT;$env:COMPILER;$env:Path"
   - sh: |
-      brew install qt5 pkg-config jack p7zip
+      export HOMEBREW_NO_INSTALL_CLEANUP=1
+      brew update
+      brew install qt@5 pkg-config jack p7zip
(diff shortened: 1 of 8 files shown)


OPNA2608:
Bump in Nixpkgs finally got merged. Overriding `patches` array because on Nix we build 0.4.6 with 0fc96c60c7ae6c2504ee696bb7dec979ac19717d added to fix build problems, but we don't need this as upstream :).

Second commit adds proper detection of PR builds in GHA, it used to always hit the `Test (unknown)` case. Third commit should fix Homebrew's download problem on AppVeyor? *I think?* We'll see. :grinning:

OPNA2608:
Considering the other commits, I decided to rename the PR to better reflect that it brings some CI improvements in general.

rerrahkr:
Thanks for your contribution!

PR #336, opened 2021-02-18
Noise distortion fix for nuked-opn-mod

### LINKED ISSUE #255: [Nuked OPN-Mod] - distortion noise on FM tones with long release rates
> #### Checklist
> 
> 
> 
> - [X] I am reporting exactly 1 bug with this issue.
> - [X] This bug hasn't already been reported.
> - [X] This bug hasn't already been fixed in the latest development build.

---

## Bug Description
When FM sound decays, some faint noise can be heard (under certain conditions, it can be very prominent) It is likely a leftover from Nuked-OPN2 - apparently a small distortion caused by the 9-bit PCM output that was post-converted into 16-bits.
How does it relate to real YM2608, I don't know, would it be possible to check whether such noise occurs on real chip as well?
It might be similiar case to #187.

## How to reproduce
_duh_
1. Create an FM instrument. Preferably, ALG=7, TL=0 for all operators and RR=1/2/3
2. Play a note

jpcima:
It drops 1 bit of quantization instead of 5, as suggested by Nuke.YKT
Fixes #255

### DIFF
--- a/BambooTracker/chip/nuked/ym3438.c
+++ b/BambooTracker/chip/nuked/ym3438.c
@@ -39,7 +39,7 @@
 #include "ym3438.h"
 
 /*OPN-MOD: define the quantization in bits at which channels clip*/
-static const unsigned channel_clipbits = 11; /*YM2612 has 9 bits*/
+static const unsigned channel_clipbits = 16; /*YM2612 has 9 bits*/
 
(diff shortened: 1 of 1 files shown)


rerrahkr:
Thanks for your contribution!

PR #342, opened 2021-03-08
Use Qt's loading/saving file dialog if OS is Linux or BSD

### LINKED ISSUE #338: [GTK-Linux|BSD] - File importer doesn't see files with upper-case extensions
> #### Checklist
> 
> 
> 
> - [X] I am reporting exactly 1 bug with this issue.
> - [X] This bug hasn't already been reported.
> - [X] This bug hasn't already been fixed in the latest development build.

---

## Bug Description

~This bug is similiar to #317 in that it happens only on specific Linux desktop environments, making it very hard to debug. I only know GNOME and XFCE sufer from that~. Steps to reproduce:
1. Right click and select "Import from bank format"
2. Navigate to directory when files in *.PVI, *.PZI or .*FF formats are (it turns out it's upper-case extension files that don't load, *.pvi for example is OK)
3. You won't see them. In order to be able to import then you must drag and drop a bank into BT's import window, and then you will see files of _every_ format, including executables, images, music files, etc.
Zrzut ekranu z 2021-02-19 11-19-44

rerrahkr:
Close #338 

Add `QFileDialog::DontUseNativeDialog` in the compilation on Linux and BSD.

QtGlobal has `Q_OS_BSD4` for BSD 4.4 system, which covers FreeBSD, NetBSD and OpenBSD iiuc. I conditioned as `defined(Q_OS_BSD4) && !defined(Q_OS_DARWIN)` to exclude macOS.

### DIFF
--- a/BambooTracker/gui/configuration_dialog.cpp
+++ b/BambooTracker/gui/configuration_dialog.cpp
@@ -665,7 +665,11 @@ void ConfigurationDialog::on_colorLoadPushButton_clicked()
 {
 	QString file = QFileDialog::getOpenFileName(this, tr("Open color scheme"),
 												QApplication::applicationDirPath() + "/skins",
-												tr("ini file (*.ini)"));
+												tr("ini file (*.ini)"), nullptr
+#if defined(Q_OS_LINUX) || (defined(Q_OS_BSD4) && !defined(Q_OS_DARWIN))
+												, QFileDialog::DontUseNativeDialog
+#endif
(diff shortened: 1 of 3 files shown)


OPNA2608:
@freq-mod Could you check if this workaround fixes the file visibility problem for you?

freq-mod:
@OPNA2608, @rerrahkr now everything is visible and loadable, thanks!

PR #353, opened 2021-04-07
Move cross-project variable logic to include

### LINKED ISSUE #352: [Windows|macOS|Linux|BSD] - PREFIX handling in wrong pro file
> #### Checklist
> 
> 
> 

---

## Bug Description

If the default `PREFIX` is used, then it's a project-local variable. When QMake reaches the `/data` subproject, the variable will have been "reset" to empty again. ~~Checking / setting of `PREFIX` probably needs to be moved to `/Project.pro`.~~ Passing variables to subprojects seems more complex that just setting `FOO = BAR` in `/Project.pro` and referencing `$$FOO` in `/data/data.pro`. I'll move it into a new pri file and include it from the subprojects that need it, probably create a new subdirectory for qmake stuff in general as well.

I'll work on this myself and open a PR soon.

## How to reproduce

1. `qmake` without a explicit `PREFIX=...`
2. `make install`
3. Note: `BambooTracker */usr/local*/bin/BambooTracker`, but `licenses **/share/doc/BambooTracker/licenses`
unknown

## System Information

- **Operating System**: Linux Mint
- **BambooTracker Version**: 0.4.6
- **Build Type**: Manual

OPNA2608:
Also add a directory for all qmake includes & tests

Closes #352 

Moved the logic that sets the `PREFIX` default and identifies the current compiler into a new `variables.pri` file so subprojects can include it if needed. Required for propagating the aforementioned `PREFIX` fallback value to the `/data/data.pro` subproject.

I took the liberty to make a new `/qmake` directory and move anything not directly related to a project in it. It now has `variables.pri` for `PREFIX` & compiler identification (BambooTracker and data), `library.pri` for building a static library and qmake tests (RtAudio and RtMidi).


rerrahkr:
Thanks for your contribution!

PR #373, opened 2021-06-15
Bump Qt versions on Windows CI

OPNA2608:
- Qt6.0.2 -> Qt6.1.1
- Qt5.15.1 -> Qt5.15.2

Qt6 bump was tested on Discord. (download link: https://github.com/OPNA2608/BambooTracker/releases/download/v0.0.0/BambooTracker-v0.0.0-windows-10-64bit.zip)
Qt5 bump is untested, but it should only be bugfixes & therefore non-breaking.

### DIFF
--- a/.github/workflows/windows-10-qt6.yml
+++ b/.github/workflows/windows-10-qt6.yml
@@ -29,7 +29,7 @@ env:
   MINGW_BITNESS: 64
   MINGW_CHOCOBASE: /c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64
   MINGW_INSTALLBASE: ${{ github.workspace }}/mingw64/
-  QT_VERSION: 6.0.2
+  QT_VERSION: 6.1.1
   QT_TOOLCHAIN: win64_mingw81
   QT_MODULES: qtbase qttools qttranslations qt5compat
   QT_INSTALLBASE: ${{ github.workspace }}/Qt/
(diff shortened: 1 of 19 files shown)


rerrahkr:
In my environment :

1. Some icons and images drawing by QPainter are oozing.
2. Some of the frames such as lists are not displayed. Sometimes resizing the window fixes it.
3. Unnecessary Hint icon "? is displayed in the title bar of many windows.

As for 3, it is probably because the default value of `QWidget::windowFlags()` was changed in Qt6. You need to change the XOR in `QWidget::setWindowFlags()` to NAND.

```c++
setWindowFlags(windowFlags() ^ Qt::WindowContextHelpButtonHint);
// Fixed
setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
```

I don't know if 1 and 2 are caused by a problem with Qt6 or my environment.

pic

rerrahkr:
> 1. Some icons and images drawing by QPainter are oozing.
> 2. Some of the frames such as lists are not displayed. Sometimes resizing the window fixes it.

They seems to be caused by Qt 6, maybe not by ours code, so there's nothing we can do to fix that at the moment. Other than that, everything is fine. I'll merge this PR and let's see what happens about the "known problem". Thanks for your contribution!

OPNA2608:
> 1. Some icons and images drawing by QPainter are oozing.

To have a visual comparison:

Before:
grafik

After:
grafik

Maybe SVG icons could be used instead, those should scale nicely with whatever size the menubar ends up being. But this may require a different / custom icon theme since Silk Icons is deliberately PNG-only.

rerrahkr:
Not only those, but also the "FM" and "SSG" icons in the instrument list are bitmaps, so it would be better to replace them with vector icons in the future.

OPNA2608:
> 1. Some icons and images drawing by QPainter are oozing.

Opened https://github.com/BambooTracker/BambooTracker/issues/402 to track this.

PR #374, opened 2021-06-20
Add compilation settings for SCCI and C86CTL

### LINKED ISSUE #357: SCCI licensing
Hello,

I looked at the SCCI license but it was in Japanese, so I gave it to DeepL for a rough translation. Here is the result:

```
About scci.dll license

This software is freeware.
There are no restrictions on the use or redistribution of this DLL or the accompanying source code for development
There are no restrictions on the use and redistribution of this DLL and the accompanying development source code, including commercial use.
You are free to use this software at your own risk.

The user is responsible for any damage to data or equipment caused by using this DLL.
The user is responsible for any damage to data or equipment caused by using this DLL.

The source code of the DLL itself is for the purpose of interface integration.
As for the source code of the DLL itself, since the purpose is to integrate the interface
The source code of the DLL itself is for the purpose of interface integration, so we will consider providing it if you contact us with your feedback.

That's all.
```

It seems that this license could be considered a custom permissive license however the legal language does not seem clear enough. In any case, it's not FSF or OSI approved and this means some distros cannot distribute it.

I actually made an openSUSE package for bambootracker. openSUSE has a [policy](https://en.opensuse.org/openSUSE:Packaging_guidelines#Licensing) which allows only FLOSS software in the community repos. If SCCI is nonfree software, that means I can only make a package for bambootracker:
1. If SCCI (including its source code) is relicensed to a license accepted by the OSI
2. If I can build BambooTracker without SCCI

I would really like to distribute BambooTracker from the music repository I help run, GeekosDAW, as I consider it a must-have for chiptune artists. 

Is there any way to solve this issue by contacting the author for relicensing or by enabling working SCCI-less builds?

rerrahkr:
Close #357.

Setting `CONFIG += real_chip`, SCCI and C86CTL sources and their wrapper classes are added to the build source list.
I have rewritten BambooTracker.pro to add this setting on Windows, and remove it on macOS and Linux by default.
If you want to use these interfaces on the latter platforms (which is unlikely), change the options in qmake.

### DIFF
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -46,7 +46,6 @@ QMAKE_CFLAGS_WARN_ON += $$CPP_WARNING_FLAGS
 QMAKE_CXXFLAGS_WARN_ON += $$CPP_WARNING_FLAGS
 
 SOURCES += \
-    chip/c86ctl/c86ctl_wrapper.cpp \
     chip/register_write_logger.cpp \
     command/instrument/swap_instruments_command.cpp \
     command/pattern/change_values_in_pattern_command.cpp \
@@ -216,11 +215,8 @@ SOURCES += \
 HEADERS += \
     bamboo_tracker_defs.hpp \
     chip/codec/ymb_codec.hpp \
-    chip/c86ctl/c86ctl.h \
-    chip/c86ctl/c86ctl_wrapper.hpp \
+    chip/real_chip_interface.hpp \
     chip/register_write_logger.hpp \
-    chip/scci/SCCIDefines.hpp \
-    chip/scci/scci.hpp \
     command/command_id.hpp \
     command/instrument/swap_instruments_command.hpp \
     command/pattern/change_values_in_pattern_command.hpp \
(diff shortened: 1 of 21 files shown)


OPNA2608:
>  If you want to use these interfaces on the latter platforms (which is unlikely), change the options in qmake.

C86CTL won't compile on anything other than Windows because it uses a Windows-specific header - the code would need to be modified to compile on non-Windows platforms. This should be fine as long as we document that the `real_chip` option is usually only compatible with Windows.

Please make sure to document the `CONFIG` value `real_chip` in the READMEs, section `Compilation`. My idea for the English version:

>   - real_chip: Compile with support for SCCI and C86CTL interfaces to a real OPNA chip<sup>4</sup>. (Default on Windows)
> 
> …
>
> <sup>4</sup>: Builds code that doesn't use an OSI-certified license. Only supported on Windows.

---

It seems like the configuration window has SCCI and C86CTL even if the build was compiled without the `real_chip` support. Could those options be excluded from the list when there's no support for them?

rerrahkr:
I modified the selection in the configuration dialog, how is this?

OPNA2608 (approved):
> I modified the selection in the configuration dialog, how is this?

Looks good to me, :+1:!

PR #382, opened 2021-07-03
Update MAME emulation

rerrahkr:
The goal of this PR is to avoid violating the license of the MAME emulation file and replace it with a new one.

Currently BT used a MAME emulation modified by VGMPlay, which seems to violate the license between the old MAME license and GPLv2. They are `BambooTracker/chip/chipdef.h` and `BambooTracker/chip/mame/*`.

The emulation used in [libvgm](https://github.com/ValleyBell/libvgm/tree/7f84cd7504b74d455edcc984aa1cd5ec580fad47) is from a newer version of MAME, and is under an OSI-approved license. Searching with MAME, the license for each of the files that could be replaced is:

- `fmopn.h/c`, `fmopn_2608rom.h`: GPL v2+
- `ymdelta.h/c`: GPL v2+
- `emu2149.h/c`: MIT
- Other files: BSD 3-Clause

The emulation will be replaced with appropriate modifications. This change may possibly affect the audio stream in some way.

---

First, I made the new emulation interface classes. The nuked emulation has also been modified for consistency.


rerrahkr:
Replaced all MAME emulations with libvgm ones. emu2149 is now managed as a submodule since it has a separate repository.

The removal of the VGMPlay-derived code means that the "GPL v2 only" files are gone. So I change the license of BT to "GPL v2 or later".
I've also modified the text in the about dialog, so those translations will need to be updated after this PR is merged.

OPNA2608:
> The emulation will be replaced with appropriate modifications. This change may possibly affect the audio stream in some way.

I'm hearing differences in No Time To Waste.btm, specifically the SSG-EG.

[No Time To Waste - FM4 Comparison.zip](https://github.com/BambooTracker/BambooTracker/files/6775758/No.Time.To.Waste.-.FM4.Comparison.zip)

I think it sounds closer to the Nuked core now, but different from what seemed to be the intended (old MAME) sound: https://www.youtube.com/watch?v=toDVpk1VOrQ

Maybe we should check which demo modules sound drastically different after these commits and ask their authors if they want to rework their modules, or if they'd like to have them removed due to the updated emulation not representing their original vision anymore.

.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "submodules/RtMidi/src"]
 	path = submodules/RtMidi/src
 	url = https://github.com/thestk/rtmidi
+[submodule "BambooTracker/chip/emu2149"]
+	path = BambooTracker/chip/emu2149
+	url = https://github.com/rerrahkr/emu2149.git
+	branch = opna-ssg
```
OPNA2608, line 10: Can this be moved to `/submodules/emu2149` instead so all submodules are consistently in one place?

This will also require a small change to the readmes[1] (and maybe a mention in the update notes): Currently you can clone the repository / download an archive of it from the GitHub UI without any submodules and `.git` data if you build against the system's Rt libraries - which brings with it an improvement in download size and purity. This new module will now require the fetching of submodules / manual cloning of [rerrahkr/emu2149](https://github.com/rerrahkr/emu2149).

[1]: From the English readme:
> ```sh
> # Can leave away --recurse-submodules if CONFIG+=system_rtaudio AND CONFIG+=system_rtmidi will be used
> git clone https://github.com/BambooTracker/BambooTracker --recurse-submodules
> ```

rerrahkr:
There seems to be a symbolic link error in appveyor.
Adding `git config core.symlinks true` to the ci job should fix it.

Or should I modify the include path of BambooTracker.pro without changing the git config?

OPNA2608:
I don't think changing local git settings to a non-default value is a good idea, a user might run into this exact build error and report it as a bug later on.

We could remove the symlink and build emu2149 as a static library instead, similar to our versions of RtAudio & RtMidi.

OPNA2608:
Building it as a static library now, AppVeyor should pass.

rerrahkr:
I sometimes hear noises when an ADPCM sample is finished playing. It seems to be caused by the memory being filled with 0xFF. I will fix it to be zero-initialized.

rerrahkr:
It works well as far as I can see. I will organize them after the merge of this PR because I have been in contact with every authors of demo modules who use SSG-EG. Thanks for your help!

PR #385, opened 2021-07-08
Fix bug where Fxx speed and tempo effects slow down playback

### LINKED ISSUE #376: Changing tempo with Fxx effect slows down song every time you use it
> #### Checklist
> 
> 
> 

---

## Bug Description

Every time you use Fxx to change a song's tempo mid-song, it resets the "partial" progress into the current row. So placing a Fxx on every row effectively rounds the song down to the next slower speed.

(Oxx groove effects do not cause a delay.)

## How to reproduce

1. Create an empty document, and add regularly spaced notes to act as a metronome.
2. Either place a single Fxx effect at the top with xx >= 20, or place Fxx effects on every row with the same xx value.
3. Compare their tempos.

For example, F80 (128 bpm) is just slightly slower than speed 7 (128.57 bpm), but if you spam F80 every row, you get speed 8 instead (112.5 bpm). The effect is more dramatic at higher tempos; for example, my example module uses FFF, which has a clearly audible change in tempo when spamming Fxx.

Example module: [bambootracker fxx tempo bug.btm.zip](https://github.com/rerrahkr/BambooTracker/files/6684020/bambootracker.fxx.tempo.bug.btm.zip)

Screenshot_20210620_212121

## System Information

- **Operating System**: Arch Linux
- **BambooTracker Version**: unstable-0142df4c01401c5ea525b68e9568fd3fd4063392
- **Build Type**: AUR build from source

### LINKED ISSUE #383: Changing *speed* with Fxx effect slows down song every time you use it
> #### Checklist
> 
> 
> 

---

## Bug Description

If you spam a F03 effect on each row of a module, it acts like Speed 4 rather than Speed 3. Related to #376.

## How to reproduce

1. Create an empty document, and add regularly spaced notes to act as a metronome.
2. Set the song speed to 3 and observe the playback rate.
3. Add a F03 effect on every row and observe the playback rate.

Sample module at [bambootracker fxx speed bug.btm.zip](https://github.com/BambooTracker/BambooTracker/files/6780354/bambootracker.fxx.speed.bug.btm.zip). This comes with two songs, one without Fxx and one with.

This occurs because at speed 3, [`TickCounter::countUp()`] returns with `restTickToNextStep_` set to 2 (due to `--restTickToNextStep_`), but a subsequent Fxx effect calls `TickCounter::setSpeed(int speed)` which calls `TickCounter::resetRest()`, which recomputes `restTickToNextStep_` without decrementing it.

[`TickCounter::countUp()`]: https://github.com/BambooTracker/BambooTracker/blob/2ac8e4c9129c732d2b5d58411ea3c91970308e24/BambooTracker/tick_counter.cpp#L105

I'm working on a fix locally which will fix both #376 and this issue.

## System Information

- **Operating System**: Arch Linux
- **BambooTracker Version**: unstable-aa4efb0e
- **Build Type**: both AUR self-built from source

nyanpasu64:
Code review walkthrough (pointing out non-obvious reasoning, arguable decisions, questions, and further issues I found):

## Fixing one-tick delay

I changed `resetRest()` to decrement `restTickToNextStep_` when the song is playing, to fix the one-tick delay issue (#383). As a result, `TickCounter::setGrooveState(GrooveState::ValidByLocal)` now only decrements restTickToNextStep_ if the song is playing. If the song is not playing (eg. "retrieve channel state"), it doesn't do so (whereas it used to). I don't *think* this has any observable effect. I didn't find any uses of `ValidByLocal` outside of playback and "retrieve channel state", but it's possible I missed something.

Also, `TickCounter::setGrooveState(GrooveState::ValidByGlobal)` (adjusting the song-global speed spinbox) now decrements restTickToNextStep_ while the song is playing (it used to never). I don't know if it makes a significant difference.

**BUG:** Both the GUI thread (changing speed spinbox) and audio thread (Fxx effect) call `TickCounter::setSpeed` without synchronization (the audio thread `PlaybackManager::streamCountUp()` acquires `PlaybackManager::mutex_` but the UI `QObject::connect(ui->speedSpinBox` doesn't). Additionally the GUI thread calling `TickCounter::setSpeed` can collide with the audio thread calling other TickCounter methods. This is a data race with undefined behavior, and can lead to unpredictable bugs.

## Fixing tempo delay

I also changed `resetRest()` to compute `tickDiffSum_` based off a cached `prevTickDiffSum_` from the most recent `countUp()` call, so that multiple calls to `resetRest()` on the same row will always produce the same result even if a tempo is active (unless a groove is active). I hope I properly assigned and reset `prevTickDiffSum_`.

To preserve the tempo state across `Fxx` tempo effects, I changed `setTempo()` to not set `tickDiffSum_` to 0. In my testing, this didn't produce issues (like rows taking too long or too short of a time) when switching between fast and slow tempos, and didn't sound wrong in my own music either. I didn't think about the code long enough to see if my changes could lead to incorrectly calculated `restTickToNextStep_`, but I didn't notice any issues using the program, and `if (restTickToNextStep_ < 1) restTickToNextStep_ = 1;` should prevent bad things from happening.

I also changed `setSpeed()` to not set `tickDiffSum_` to 0 (a debatable choice). This is because in BambooTracker and FamiTracker, changing speed doesn't cause previously set tempo to stop taking effect, so I decided to not reset the tempo state either. However this has the downside that you can't reset the tempo accumulator mid-song using a speed command.

- Perhaps setGrooveState should reset `tickDiffSum_` and `prevTickDiffSum_`, since in BambooTracker, groove and tempo are mutually exclusive. I didn't make this change, but I could add it.

## Misc

I explicitly initialized all fields in `TickCounter::TickCounter()`. I wanted to initialize `prevTickDiffSum_` upfront because I didn't know when it would otherwise be initialized, and decided to initialize the other fields as well. I don't like leaving uninitialized fields around, since they're a landmine that creates problems if you later unknowingly change the code to read an uninitialized variable.

- I actually don't like the look of the initializer list I wrote, since it's both verbose and easy to forget to initialize fields. I'd kinda prefer to put the initializers in the header file (like `bool isPlaySong_ = false`), but decided against it since it didn't make sense to put `tempo_ = 150` in the header file. Honestly I think the presence of a default tempo, tick rate, and step size at all is a code smell; the initial values should be explicitly passed into the constructor. I don't like how the default tempo of 150 is duplicated in the .ui file, song.hpp, *and* tick_counter.cpp.

----

An earlier draft of this PR introduced a new variable `shouldTickRest_` set to true in `countUp()` and false in `resetCount()`, and read by `resetRest()`. However I messed it up; it should only be set to true if `isPlaySong_` was true, and it introduced another moving part that I might've failed to set and unset at the right times. I decided to remove `shouldTickRest_` altogether and use `isPlaySong_` as a substitute. This is simpler, but I don't know if it's correct either. **Will this cause any problems?**

**Question:** Why does `GrooveState::ValidByGlobal` initialize nextGroovePos_ to size - 1? Is it OK if I change both GrooveState values to decrement `restTickToNextStep_` if the song is playing?

----

(Sidenote) I'm having some trouble reasoning about this code, the order of events across multiple files, and I don't know what methods are called in what order, but I do know that the order affects how the code acts. I feel ["John Carmack on Inlined Code"][1] is a good argument against tangled codebases with deeply nested control flow.

[1]: http://number-none.com/blow/john_carmack_on_inlined_code.html

Fixes #376.
Fixes #383.

### DIFF
--- a/BambooTracker/tick_counter.cpp
+++ b/BambooTracker/tick_counter.cpp
@@ -138,11 +138,13 @@ void TickCounter::resetRest()
 		tickDiffSum_ += tickDiff_;
 		int castedTickDifSum = static_cast<int>(tickDiffSum_);
 		restTickToNextStep_ = defStepSize_ + castedTickDifSum;
-		if (restTickToNextStep_ < 1) restTickToNextStep_ = 1;	// Prevent wait count changing to 0 (freeze)
 		tickDiffSum_ -= castedTickDifSum;
 	}
 	else {
 		restTickToNextStep_ = grooves_.at(static_cast<size_t>(nextGroovePos_));
 		nextGroovePos_ = (nextGroovePos_ + 1) % static_cast<int>(grooves_.size());
 	}
+
+	// Ensure each row lasts for at least 1 tick, and that we can subtract 1 safely.
+	if (restTickToNextStep_ < 1) restTickToNextStep_ = 1;
 }
--- a/BambooTracker/tick_counter.cpp
+++ b/BambooTracker/tick_counter.cpp
@@ -92,6 +92,12 @@ void TickCounter::setGrooveState(GrooveState state)
 		resetRest();
 		break;
 	}
+
+	// When enabling groove (which disables tempo), reset tempo accumulator.
+	if (state == GrooveState::ValidByGlobal || state == GrooveState::ValidByLocal) {
+		tickDiffSum_ = 0.f;
+		prevTickDiffSum_ = 0.f;
+	}
 }
 
 bool TickCounter::getGrooveEnabled() const noexcept
--- a/BambooTracker/tick_counter.cpp
+++ b/BambooTracker/tick_counter.cpp
@@ -83,7 +83,6 @@ void TickCounter::setGrooveState(GrooveState state)
 	case GrooveState::ValidByLocal:
 		nextGroovePos_ = 0;
 		resetRest();
-		--restTickToNextStep_;	// Count down by step head
 		break;
 	case GrooveState::Invalid:
 		nextGroovePos_ = -1;
@@ -110,8 +109,9 @@ int TickCounter::countUp()
 		if (!restTickToNextStep_) {  // When head of step, calculate real step size
 			resetRest();
 		}
-
-		--restTickToNextStep_;   // Count down to next step
+		else {
+			--restTickToNextStep_;	 // Count down to next step
+		}
 
 		return ret;
 	}
@@ -147,4 +147,10 @@ void TickCounter::resetRest()
 
 	// Ensure each row lasts for at least 1 tick, and that we can subtract 1 safely.
 	if (restTickToNextStep_ < 1) restTickToNextStep_ = 1;
+
+	// If resetRest() is called in response to the tracker processing a row or Fxx/Oxx event,
+	// subtract 1 so countUp() will remain on the row for (speed - 1) subsequent ticks.
+	if (isPlaySong_) {
+		restTickToNextStep_--;
+	}
 }
--- a/BambooTracker/tick_counter.cpp
+++ b/BambooTracker/tick_counter.cpp
@@ -80,8 +80,8 @@ void TickCounter::setGrooveState(GrooveState state)
 {
 	switch (state) {
 	case GrooveState::ValidByGlobal:
-		nextGroovePos_ = static_cast<int>(grooves_.size()) - 1;
-		resetRest();
+		nextGroovePos_ = 0;
+		if (isPlaySong_) restTickToNextStep_ = grooves_.at(0) - 1;
 		break;
 	case GrooveState::ValidByLocal:
 		nextGroovePos_ = 0;
--- a/BambooTracker/tick_counter.cpp
+++ b/BambooTracker/tick_counter.cpp
@@ -30,7 +30,11 @@ TickCounter::TickCounter() :
 	tempo_(150),    // Dummy set
 	tickRate_(60),	// NTSC
 	nextGroovePos_(-1),
-	defStepSize_(6)    // Dummy set
+	defStepSize_(6),	// Dummy set
+	restTickToNextStep_(0),
+	tickDiff_(0.f),
+	tickDiffSum_(0.f),
+	prevTickDiffSum_(0.f)
 {
 	updateTickDifference();
 }
@@ -45,7 +49,6 @@ void TickCounter::setTempo(int tempo)
 {
 	tempo_ = tempo;
 	updateTickDifference();
-	tickDiffSum_ = 0;
 	resetRest();
 }
 
@@ -58,7 +61,7 @@ void TickCounter::setSpeed(int speed)
 {
 	defStepSize_ = speed;
 	updateTickDifference();
-	tickDiffSum_ = 0;
+	// Changing speed does not reset tempo, so don't reset tickDiffSum_ either.
 	resetRest();
 }
 
@@ -103,6 +106,7 @@ void TickCounter::setPlayState(bool isPlaySong) noexcept
 
 int TickCounter::countUp()
 {
+	prevTickDiffSum_ = tickDiffSum_;
 	if (isPlaySong_) {
 		int ret = restTickToNextStep_;
 
@@ -130,12 +134,14 @@ void TickCounter::resetCount()
 {
 	restTickToNextStep_ = 0;
 	tickDiffSum_ = 0;
+	prevTickDiffSum_ = 0;
 }
 
+// MUST be idempotent when no groove is active.
 void TickCounter::resetRest()
 {
 	if (nextGroovePos_ == -1) {
-		tickDiffSum_ += tickDiff_;
+		tickDiffSum_ = prevTickDiffSum_ + tickDiff_;
 		int castedTickDifSum = static_cast<int>(tickDiffSum_);
 		restTickToNextStep_ = defStepSize_ + castedTickDifSum;
 		tickDiffSum_ -= castedTickDifSum;
--- a/BambooTracker/tick_counter.hpp
+++ b/BambooTracker/tick_counter.hpp
@@ -68,6 +68,7 @@ private:
 
 	float tickDiff_;
 	float tickDiffSum_;
+	float prevTickDiffSum_;
 
 	void updateTickDifference();
 	void resetRest();


BambooTracker/tick_counter.cpp
```diff
@@ -58,7 +61,12 @@ void TickCounter::setSpeed(int speed)
 {
 	defStepSize_ = speed;
 	updateTickDifference();
-	tickDiffSum_ = 0;
+	// I choose to not set tickDiffSum_ to 0.
+	// This is because in BambooTracker and FamiTracker,
+	// changing speed does not reset tempo,
+	// so I decided to not interfere with the tempo accumulator either.
+	// However not resetting tickDiffSum_ has the downside that
+	// you can't reset the tempo accumulator mid-song using a speed command.
```
OPNA2608, line 69: Explaining your rationale behind a change is better suited to the commit / PR description than a comment within a function.
```suggestion
```

OPNA2608, line 69: Does this technique have a real-world use in a module or is this a strictly technical observation of behind-the-scenes behaviour? Have you seen it being used anywhere - and would this PR therefore break any existing modules?

BambooTracker/tick_counter.cpp
```diff
@@ -130,19 +139,51 @@ void TickCounter::resetCount()
...
+	Currently, resetRest() decrements restTickToNextStep_ if the song is playing.
+
+	There are other cases when resetRest() is never called (I think) unless you change
+	global settings:
+
+	- Before countUp() processes head of step (the current code always calls countUp()
+	  before calling resetRest())
+	- After countUp() processes middle of step (never happens, Fxx effects are always
+	  processed at the head, not delayed by a Gxx effect)
+	*/
```
OPNA2608, line 185: This whole comment feels very excessive for a simple `if (something == true) then decrement(thatThing) end`. Some of that might be more appropriate in a commit description or function-level comment?

OPNA2608 (commented):
Some comments on your comments, all IMO.

BambooTracker/tick_counter.cpp
nyanpasu64: Currently, every tempo effect and every speed effect instantly reset the tempo accumulator(?). And one other person and I believe that *tempo* effects causing a delay is a bug which prevents smooth tempo transitions.

I don't know if there are any modules that depend on speed effects resetting the tempo accumulator. This will only affect modules which change speed using `F0x/F1x` after the tempo is changed away from 150 (or whatever equivalent value if the tick rate is changed from 60 Hz) using global tempo or `Fxx`. Even then, it should only have a minor effect, causing notes to be played 1 tick sooner or later (per row with a Fxx) at most. (I suspect that tempos above and below 150 have a different effect, because they result in `tickDiffSum_` having different signs, and `static_cast<int>(tickDiffSum_)` rounds towards 0.)

IMO, a bigger concern is fixing how each row with a `Fxx` has a full 1-tick delay. I think this is clearly a bugfix, but it might change how modules sound.

Note: the tempo system isn't really an "accumulator" like the N-SPC/Squaresoft SPC tempo systems are. It's implemented strangely internally, where the nominal speed and the tempo offset from the nominal speed are stored separately; maybe this can be changed to be more clear. What can't be changed is how BambooTracker computes the instantaneous duration of each row upfront, in order to release notes 1 tick before a new note triggers.

BambooTracker/tick_counter.cpp
```diff
@@ -130,19 +139,51 @@ void TickCounter::resetCount()
...
 	else {
 		restTickToNextStep_ = grooves_.at(static_cast<size_t>(nextGroovePos_));
 		nextGroovePos_ = (nextGroovePos_ + 1) % static_cast<int>(grooves_.size());
 	}
+
+	// restTickToNextStep_ can be 0 in case of a groove containing 0.
+	// Previously this function wrote 0, and TickCounter::countUp() subtracted 1,
+	// resulting in restTickToNextStep_ = -1, and waiting 4 billion ticks
+	// (plus signed integer overflow UB) for it to reach 0 and advance to the next row.
+	// To avoid this, clamp non-positive values to 1 in both speed and groove.
```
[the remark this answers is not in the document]
nyanpasu64: I'll look into removing or moving comments tomorrow-ish.

OPNA2608:
Now for your walkthrough: (I'm leaving any commenting on the code side of things to Rerrah because he's more familiar with how everything fits together)

> **BUG:**

Could you open an issue about this please?

> Honestly I think the presence of a default tempo, tick rate, and step size at all is a code smell; the initial values should be explicitly passed into the constructor. I don't like how the default tempo of 150 is duplicated in the .ui file, song.hpp, and tick_counter.cpp.

This seems like a genuine codebase problem to me, especially if this affects other default values. Maybe open an issue about this as well?

> (Sidenote) I'm having some trouble reasoning about this code, the order of events across multiple files, and I don't know what methods are called in what order, but I do know that the order affects how the code acts. I feel ["John Carmack on Inlined Code"](http://number-none.com/blow/john_carmack_on_inlined_code.html) is a good argument against tangled codebases with deeply nested control flow.

I do agree with your observation and I think that the codebase in general - as someone who's had more contact with Java than C++ - has some problems that makes reading into it quite un-fun, i.e. a general lack of documentation. I think it's understandable *why* this happened, and having *a* cross-platform YM2608 tracker with these flaws is certainly better than having none at all. Beggars can't be choosers and all :sweat_smile:.

Rerrah has announced the following on the Discord server afew weeks ago:

> The reason I am considering [creating a BambooTracker GitHub organisation] is that I am no longer as enthusiastic about Bamboo development as I used to be.
> As of last summer, Bamboo had almost enough functionality for me.

> I don't have a clear roadmap on how to go about future development.
> So please feel free to edit the repository, wiki, and website as you see fit.

With this in mind, I'm not sure how much Rerrah himself wants to do about these codebase observations. None of the current other members of the organisation work in C++ (I do Java full-time and CI/QMake/general structure stuff in BT, freq-mod does translation work). But I'm happy to accept refactoring & function-documenting PRs if you (or someone else) have the motivation to do it!

BambooTracker/tick_counter.cpp
OPNA2608: I see, thanks for the thorough explanation.

nyanpasu64: I'm wondering if it's better to *keep* resetting the tempo system when the user inputs a speed effect. The benefit is that it allows the user to reset tempo phase (though you technically can already do so using a groove), which is helpful if they decide to repeatedly switch between tempo 150 and another value (a contrived scenario). The downside is you can't remain at a custom tempo but switch between speed 6 for some rows, and speed 3 for other rows, while maintaining tempo rhythm (also a contrived scenario)... In other words, I don't really know.

nyanpasu64:
Is there a problem if `resetRest()` unconditionally decrements `restTickToNextStep_`? Will the value of `restTickToNextStep_` while the song isn't playing actually get observed in any case?

One issue with branching on `isPlaySong_` is that initiating playback produces different `restTickToNextStep_` depending on whether you play from stop, or initiate playback while already playing. However this difference is unobservable because `PlaybackManager::playStep()` and `PlaybackManager::startPlay()` subsequently call `resetCount()` which sets restTickToNextStep_ to 0. I'm not sure if there's any other behavioral deviations I've introduced.

nyanpasu64:
Force-pushed, for better or for worse. I'm no longer convinced that my patches are "correct" and follow the intent of the code, but it fixes a bug and I haven't noticed any issues with my changes *so far*.

rerrahkr (commented):
As far as I can see, it works fine.

I agree that `tickDiffSum_` and `prevTickDiffSum_` should be cleared when groove is set.

When the song is not playing, tempo, step, and groove are not functioning. When the song is stopped, even if `countUp()` is called, it will only return -1 and is not supposed to observe `restTickToNextStep_`. so whatever the value of `restTickToNextStep_` is, it should have no effect. I think it would be better to just branch out, so that you don't have to make unnecessary changes.

---

Another annoyance is that the bug that existed before has been fixed, and now the song is playing at a different speed than before. In particular, if you change the step size frequently, you will need to do the following to get the song to play at the same speed as before.

```
| 0F05 |           | 0F06 |
| ---- |           | 0F05 |
| 0F03 |           | 0F04 |
| ---- |           | 0F03 |
| 0F05 |    ==>    | 0F06 |
| ---- |           | 0F05 |
| 0F03 |           | 0F04 |
| ---- |           | 0F03 |
...
```

We need to announce this change in IMPORTANT.md, Discord, twitter, etc.

nyanpasu64:
> I think it would be better to just branch out, so that you don't have to make unnecessary changes.

@rerrahkr What did you mean? Do you mean I should always decrement `restTickToNextStep_` rather than only when playing?

rerrahkr:
It only decrements during playback, it's not a problem.
Decrementing when not playing is meaningless. `if (isPlaySong_) { ... }` can make it explicit.

nyanpasu64:
I tried not editing resetRest() and instead branching in every call to that method. Is this what you wanted me to try? https://github.com/nyanpasu64/BambooTracker/commit/11bcaef0730ad09930466adf58d686a0a31c6dd6

Personally I don't like the duplicated logic, and think it would be easy to forget to perform this test when adding a new effect or refactoring.

rerrahkr:
I misunderstand the meaning of "branch" properly. I think it's better to decrement inside `resetRest()` like this PR. I'm sorry to have troubled you.

rerrahkr:
> Why does `GrooveState::ValidByGlobal` initialize nextGroovePos_ to size - 1? Is it OK if I change both GrooveState values to decrement `restTickToNextStep_` if the song is playing?

https://github.com/nyanpasu64/BambooTracker/blob/9bd34ca06fd44726cbbb91336b3058dcf4e9501e/BambooTracker/tick_counter.cpp#L79-L85

The reason is to set `nextGroovePos_` to 0 in `resetRest()`. The value of the last groove sequence is stored in `resetTickToNextStep_` at the same time, but it has no effect since `resetCount()` is called when playback starts. When the first step is loaded, `resetRest()` is called again, and the appropriate values are set to them.

Now that I've followed the behavior again, I feel like I'm forcing `ValidByGlobal` process into `resetRest()`, which is hard to understand. Is it better to write a simpler process inline instead of forcing it?

Also, when `GrooveState::ValidByGlobal` is set during playback, `restTickToNextStep_` is set to the value of the last groove sequence, which may make trouble.

nyanpasu64:
I don't know what you want to do. Do you mean `GrooveState::ValidByGlobal` shouldn't call `resetRest`, but instead reset `nextGroovePos_` to 0 by hand (and set `restTickToNextStep_` to some value but I don't know which)? I don't know what value to set, you mentioned `restTickToNextStep_` doesn't matter when the song isn't playing, but I don't know what to set it to when the song is playing... (though I think no behavior is correct, since writing to this value during playback is probably a data race.)

Alternatively, can you create a branch (or push to my branch) with your suggested changes to the code (as of either master or my branch)?

rerrahkr:
5bfcd5db20e2be6126824a7dbfcb593b7698e163: When changing the groove state during playback, set the first value of the groove sequence decremented to `restTickToNextStep_`. I think this is simpler.

nyanpasu64:
Pushed my branch with a new commit. Switching global grooves seems to work fine, testing in a document where I added 2 global grooves.

rerrahkr:
I'm ready to merge this PR. Are there any additional points you would like to modify?

nyanpasu64:
> Perhaps setGrooveState should reset `tickDiffSum_` and `prevTickDiffSum_`, since in BambooTracker, groove and tempo are mutually exclusive. I didn't make this change, but I could add it.

Should I make this change, test it, and push a new commit?

The data race when changing global speed isn't trivial to fix, and I'd defer it to a later PR. My ideas for fixing it include having the GUI acquire the mutex while changing speed (I don't know if it happens between ticks or during a tick), or adding some sort of atomic protocol for the GUI to signal when the global speed changes, and the audio thread checks it once a tick.

rerrahkr:
Thanks for your contribution!

PR #398, opened 2021-08-02
Fix MSVC support

OPNA2608:
This PR fixes support for building with MSVC (at least a recent version of it).

Downgrades the strictness of the MSVC compiler flags due to *many* problems that are either too ridiculous (undefined compiler macros -> error) or too complicated (errors in Qt's own headers, should use MSVC equivalent of `-isystem` but those includes come from Qmake spec files) to resolve nicely.

Any support for MSVC is better than no support though :smile:.

CC @djtuBIG-MaliceX

### DIFF
--- a/BambooTracker/bamboo_tracker.cpp
+++ b/BambooTracker/bamboo_tracker.cpp
@@ -27,6 +27,7 @@
 #include <algorithm>
 #include <utility>
 #include <exception>
+#include <iterator>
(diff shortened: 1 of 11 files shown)


djtuBIG-MaliceX:
Note that these changes were tested for msvc2019 given reasonably improved coverage of standards compliance.  There is no guarantee that this will work fine for older msvc compilers.

PR #404, opened 2021-08-19
Fix bug where pattern header was sometimes not redrawn

nyanpasu64:
In 0.5.0 and before, pressing Shift+Tab repeatedly until you reach FM 2 will fail to redraw the header, because `oldLeftTrackIdx` was erroneously declared a `bool`.

In builds since accb5ac5f64a82560fff775f9a36f1a4642007d6 (https://github.com/BambooTracker/BambooTracker/commit/accb5ac5f64a82560fff775f9a36f1a4642007d6#diff-97d96a2c599e40b88ae7e016d1b61a16bbb5cb81d4ef24f08c83fa5e8f2307fb) (more recent than 0.5.0), placing the cursor in the leftmost column, then pressing Shift+Tab to jump to the right an even number of columns, will fail to redraw the header, and 1 of every 2 subsequent Shift+Tab presses will fail to redraw. This is due to the bizarre `static_cast<bool>(leftTrackVisIdx_&1) != oldLeftTrackIdx` comparison.

This fixes both problems by declaring `oldLeftTrackIdx` as an `int`, and comparing it to `leftTrackVisIdx_` directly.

### DIFF
--- a/BambooTracker/gui/pattern_editor/pattern_editor_panel.cpp
+++ b/BambooTracker/gui/pattern_editor/pattern_editor_panel.cpp
@@ -1140,7 +1140,7 @@ int PatternEditorPanel::calculateColNumInRow(int trackVisIdx, int colNumInTrack,
 void PatternEditorPanel::moveCursorToRight(int n)
 {
 	int oldTrackIdx = curPos_.trackVisIdx;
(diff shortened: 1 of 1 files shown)


nyanpasu64:
not waiting for CI to pass is... brave.

OPNA2608:
Yes.

PR #405, opened 2021-08-20
Add CMake build option

### LINKED ISSUE #397: [Feature] - Alternative build system options
> #### Checklist
> 
> 
> 

---

## Requested Feature
@nyanpasu64 suggested that some problems with our QMake script (i.e. #395 or dodgyVisual Studio support) might get solved if we convert to an alternate build system. I think this is worth discussing on the issue tracker.

Some alternatives that were mentioned/discussed & my opinions on them:

- qbs

I'm not really familiar with Qbs so I can't comment on the system itself.

- cmake

I'm personally in favour of CMake so far. I've had some experience with it and I think it's an alright & powerful tool. It's also the build system Qt itself uses, promotes and actively improves integration with in Qt6.

- meson

Meson seems very popular with new developers due to its ease-of-use, but that comes at a cost of some opinionated design decisions that make it differ from more traditional build systems, from what I can tell. [My Linux distro of choice for example has been employing custom patches to fix Meson behaviour - some for more than four(!) years, some with upstream rejecting the patches (so far).](https://github.com/NixOS/nixpkgs/blame/3b4f4560d286847dab73d412567da48411db861c/pkgs/development/tools/build-managers/meson/default.nix#L19-L60)

We should also think about whether we want to fully drop QMake support or only offer a secondary build system option. This should involve checking whether the alternate system of choice has proper (and easy-to-use) support for the Qt versions we want to support (Windows XP needs ~ 5.5, and what about #247?) and if every active contributor is willing to accept a complete switch.

OPNA2608:
Closes #397.

Qt5-only WIP, only tested on Linux.


BambooTracker/CMakeLists.txt
```diff
@@ -0,0 +1,279 @@
...
+	list(APPEND BT_WARNFLAGS /W3 /WX /utf-8 /D_CRT_SECURE_NO_WARNINGS)
+else(MSVC)
+	list(APPEND BT_WARNFLAGS -Wall -Wextra -pedantic -Werror -pedantic-errors)
+endif(MSVC)
+
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_AUTORCC ON)
+set(CMAKE_AUTOUIC ON)
+
+find_package(Qt5 COMPONENTS Core Gui Widgets LinguistTools REQUIRED)
```
OPNA2608, line 20: Needs logic to handle Qt6.

BambooTracker/lang/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+qt5_add_translation(TRANSLATIONS
+	bamboo_tracker_ja.ts
+	bamboo_tracker_fr.ts
+	bamboo_tracker_pl.ts
+)
```
OPNA2608, line 5: Needs logic to handle Qt6.

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+cmake_minimum_required(VERSION 3.0)
```
OPNA2608, line 1: Dunno what a good cutting-off point would be w/r/t lowest supported CMake version, just picked the lowest 3.x version.

OPNA2608: - XP CI ships CMake 3.12.2.
- Ubuntu 18.04 ships 3.10.2.
- Can't check what Ubuntu 16.04 ships, Ubuntu removed xenial from their package search.

BambooTracker/CMakeLists.txt
```diff
@@ -0,0 +1,279 @@
...
+
+target_compile_options(BambooTracker PRIVATE ${BT_WARNFLAGS})
+
+option(REAL_CHIP "Compile with support for SCCI and C86CTL interfaces to a real OPNA chip" WIN32)
+
+if(REAL_CHIP)
+	target_compile_definitions(BambooTracker PRIVATE USE_REAL_CHIP)
+endif(REAL_CHIP)
+
+target_link_libraries(BambooTracker PUBLIC Qt5::Core Qt5::Gui Qt5::Widgets)
```
OPNA2608, line 250: Needs logic to handle Qt6.

OPNA2608:
Builds & packages on MinGW Qt6 CI now.

- Fix DirectSound API capability detection (?)
- Migrate some Qt5 CI to CMake too?

OPNA2608:
- There's no detection for DirectSound support, adding the DS enable option on CI will have to do (could set a cache variable to enable it by default but *meh*)
- Leaving the Qt5 CI on the known-working QMake config is prolly better. I've tested the CMake script on x86_64-linux and x86_64-darwin and it works well enough for a local installation (revert the last commit to test via the Nix build script), haven't looked into macOS deployment

PR #409, opened 2021-08-26
Fix instrument editor envelope graphs on Breeze master

### LINKED ISSUE #403: [Linux] - "Incorrect" method of drawing instrument editor graphs results in missing graphs on KDE Plasma Breeze dev
> #### Checklist
> 
> 
> 

---

## Bug Description

On KDE, with dev builds of the Breeze theme installed, BambooTracker's instrument editor's graphs are replaced with white rectangles.

## How to reproduce

1. Install KDE Plasma, and build and install Breeze at or after f6ef4f4738a5d3b6a07b7b11451af19635c8acd9 "Draw background for frames". (I edited the `breeze` Arch PKGBUILD to do so, maybe it's easier to fire up a KDE Neon Unstable VM which should ship with KDE master).
2. Install and run BambooTracker.
3. Create and double-click an instrument.

I reported this bug to KDE, and was told ([link](https://bugs.kde.org/show_bug.cgi?id=440855#c1)) that BambooTracker was drawing the graph incorrectly.

## System Information

- **Operating System**: Arch Linux
- **BambooTracker Version**: unstable-b91e658 (b91e658d7ba85c1c696ee065f043c9b9ad37149f)
- **Build Type**: PKGBUILD

nyanpasu64:
Previously FMOperatorTable would paint a custom graph on a QFrame. On KDE's upcoming Breeze theme revision, the QFrame overwrites the contents of the graph by filling it with a background color.

I changed the QFrame to add a placeholder child QWidget, which I made to fill the QFrame using a placeholder vertical layout with zero content margins. The child QWidget has `native="true"`, which sounds scary (in Qt, native widgets have their own HWND or the X11 equivalent, which is undesirable here), but the generated C++ code doesn't actually mention `native` at all. [Stack Overflow says](https://stackoverflow.com/questions/6257908/what-does-native-true-stand-for-in-a-qt-designer-form) `native="true"` marks the widget as using native theming, but I'm not sure about that explanation.

The alternative is to not put a child widget in the QFrame, but instead set the QFrame to have no borders. But I think that would look uglier, since the result wouldn't have a frame border.

Fixes #403.

(I have a similar commit, not in this PR yet, for surrounding the wave viewer oscilloscope with a QFrame.)

----

In Qt, what's the best way to give a custom-painted widget a border, but make it not count towards the `sizeHint()` or `width()` or the `QPainter`-accessible area? Surround it in a `QFrame`? Now the owning widget needs to insert the QFrame into the layout, but communicate with its child widget.

QGraphicsView and QAbstractItemView instead inherit from QAbstractScrollArea, and place the custom logic in the parent widget instead of the child (which is a base-class QWidget). But placing custom logic in the owner/parent which draws onto a child feels awkward to me (and I don't know how to manage it), and inheriting from QAbstractScrollArea feels unsuited for non-scrolling widgets.

### DIFF
--- a/BambooTracker/gui/instrument_editor/fm_operator_table.cpp
+++ b/BambooTracker/gui/instrument_editor/fm_operator_table.cpp
@@ -108,7 +108,8 @@ FMOperatorTable::FMOperatorTable(QWidget *parent) :
 		if (!isIgnoreEvent_) emit operatorValueChanged(Ui::FMOperatorParameter::SSGEG, value);
 	});
 
-	ui->envFrame->installEventFilter(this);
+	ui->envGraph->setBackgroundRole(QPalette::Base);
+	ui->envGraph->installEventFilter(this);
 }
 
 FMOperatorTable::~FMOperatorTable()
@@ -204,11 +205,10 @@ QString FMOperatorTable::toString() const
 
 bool FMOperatorTable::eventFilter(QObject* obj, QEvent* event)
 {
-	if (obj == ui->envFrame) {
+	if (obj == ui->envGraph) {
 		if (event->type() == QEvent::Paint) {
-			QPainter painter(ui->envFrame);
-			painter.eraseRect(ui->envFrame->rect());
-			painter.drawPixmap(ui->envFrame->rect(), envmap_, envmap_.rect());
+			QPainter painter(ui->envGraph);
+			painter.drawPixmap(ui->envGraph->rect(), envmap_, envmap_.rect());
 		}
 	}
 
@@ -229,7 +229,13 @@ void FMOperatorTable::resizeEvent(QResizeEvent*)
 
 void FMOperatorTable::resizeGraph()
 {
-	envmap_ = QPixmap(ui->envFrame->size());
+	// Resize the graph widget to fill the entire QFrame.
+	// This avoids issues (eg. on the adwaita-qt theme)
+	// where the graph is surrounded by wide margins.
+	// The edges may be cut off, but it's not an issue in practice.
+	ui->envGraph->setGeometry(ui->envFrame->rect());
+
+	envmap_ = QPixmap(ui->envGraph->size());
 	xr_ = (envmap_.width() - (ENV_LINE_W_ + 1) * 2.) / ENV_W_;
 	yr_ = (envmap_.height() - (ENV_LINE_W_ + 1) * 2.) / ENV_H_;
 }
@@ -425,7 +431,7 @@ void FMOperatorTable::repaintGraph()
 		}
 	}
 
-	ui->envFrame->repaint();
+	ui->envGraph->repaint();
 }
 
 void FMOperatorTable::on_ssgegCheckBox_stateChanged(int)
--- a/BambooTracker/gui/instrument_editor/fm_operator_table.ui
+++ b/BambooTracker/gui/instrument_editor/fm_operator_table.ui
@@ -271,6 +271,8 @@
         <property name="lineWidth">
          <number>1</number>
         </property>
+        <widget class="QWidget" name="envGraph" native="true">
+        </widget>
        </widget>
       </item>
      </layout>


OPNA2608:
Some Before - After screenshots.

---

<details>
  <summary>Linux - PantheonDE's default Adwaita GTK3 theme</summary>
  
Bildschirmfoto_2021-08-28_16-03-18

The graph receives some noticeable extra padding which I think looks better. The inherited background colour also looks nicer.
</details>

---

<details>
  <summary>Linux - Kvantum Qt5 theme (BambooTracker -style kvantum)</summary>
  
Bildschirmfoto_2021-08-28_16-32-33

I can't notice any difference here.
</details>

---

<details>
  <summary>macOS</summary>
  
Bildschirmfoto_2021-08-28_16-16-31

Here the inherited colour doesn't look nice, makes it seem to me like the graph is disabled.
</details>

---

<details>
  <summary>Platform-agnostic "Fusion" (BambooTracker -style Fusion)</summary>
  
Bildschirmfoto_2021-08-28_16-04-21
Bildschirmfoto_2021-08-28_16-18-48

Slightly more padding & inherited background colour, looks okay.
</details>

---

I think it's generally an okay change, though I'd prefer if we could set the graph's background to i.e. "TextEdit background" so it looks nice on macOS again.

nyanpasu64:
The change in padding on Elementary is weird, considering I set the layout margins to 0. I'd have to look into it more. I don't know if/why scroll areas don't have that extra padding considering they do something similar.

nyanpasu64:
I reproduced the thick borders using https://aur.archlinux.org/packages/adwaita-qt/. Interestingly, QAbstractScrollArea-based widgets (like QGraphicsView and QListWidget) don't have thick borders, and when I open GammaRay and click on the contents, it selects the parent widget rather than the contents (unlike my QFrame contraption which focuses the child widget doing the drawing). This requires further research to get right. It might be okay as a hack-fix for now, not sure, maybe I'll replicate QAbstractScrollArea without scrolling, maybe I'll just reuse it.

nyanpasu64:
> I think it's generally an okay change, though I'd prefer if we could set the graph's background to i.e. "TextEdit background" so it looks nice on macOS again.

This is possible via `setBackgroundRole(QPalette::Base);` like `QAbstractScrollAreaPrivate::init()` does. But I'm still researching how QAbstractItemView and QAbstractScrollArea get rid of the margins even on Qt Adwaita.

nyanpasu64:
The previous code ran: `painter.eraseRect(ui->envFrame->rect());` (replace with transparency or white? on adwaita-qt, it instead replaced with off-white:)

Screenshot_20210902_213017_

Then `painter.drawPixmap(ui->envFrame->rect(), envmap_, envmap_.rect());`. Does QPainter have an alpha channel? I think it's platform-dependent, being `Format_ARGB32_Premultiplied` on Linux, and `Format_RGB32` on Windows (in exotracker, didn't try BambooTracker), and I didn't test on Mac. And the presence or absence of an alpha channel affects the background color of `envmap_` produced by `envmap_.fill(palette_->instFMEnvBackColor)` (255, 255, 255, 0). I feel this kind of platform-dependence makes it harder to reason about code, and I don't even know the intended behavior at this point.

----

When testing my newest force-push on my Linux machine, changing the parameter of `setBackgroundRole(QPalette::Base)` didn't change the color of the widget, even with the event filter removed. I'm not sure if it fixes the background color on Mac, or if `instFMEnvBackColor` should be changed to either white or a platform-specific background color (which varies with the current theme?). But it's still worth testing to see if the bug is fixed.

I still think that if `instFMEnvBackColor` remains transparent, `envmap_` needs to change from a QPixmap to a QImage with `Format_ARGB32[_Premultiplied?]` to ensure consistent behavior across platforms.

nyanpasu64:
Note that my latest force push changes fm_operator_table.ui to place a QWidget inside a QFrame. I hand-edited the .ui file to not specify a geometry for the nested widget, because the position will get overriden by the .cpp code upon startup and window resize. However, future developers editing/saving the .ui file will generate spurious diffs. What's the best approach to this? Move the nested widget out of the .ui layout into a regular field?

OPNA2608:
> I hand-edited the .ui file to not specify a geometry for the nested widget […]
> However, future developers editing/saving the .ui file will generate spurious diffs.

Yeah this seems like something that should be avoided. If the position gets overridden anyway, what's the problem with leaving it in?

nyanpasu64:
> Yeah this seems like something that should be avoided. If the position gets overridden anyway, what's the problem with leaving it in?

It's mostly ideological/opinionated, where I feel that arbitrary magic numbers don't belong in source control, and the program shouldn't be imposing them, and users should be vigilant to make sure they don't make it to source control which can pollute diffs if they ever change. But it's not the path of least resistance, or the most practical one.

I noticed the parent `envFrame` has a hard-coded `minimumSize` but not geometry. Maybe it's better (sadly) to just leave the geometry in the .ui file.

Plasma 5.23 is planned to release Oct 7, 2021, and rolling-release distributions will pick it up over the following days and weeks. If a border around the graph is desired but you don't want nested widgets, you could replace `envFrame` with a `QAbstractScrollArea` subclass. The lowest-effort solution to "just make it work on KDE" is to disable the QFrame's background drawing, or replace it with a base QWidget.

I still haven't replaced `envmap_` with a QImage, nor fixed the background color... IDK when I'll get around to revisiting this.

PR #412, opened 2021-09-01
Keep instrument editors on top of main window, spawn centered

nyanpasu64:
This changes all instrument editor dialogs to QDialog subclasses parented to the main window. As a result, the dialogs remain on top of the main window and open centered within it. This matches FamiTracker and probably most user expectations (I know of myself and one other person who dislikes the current behavior).

It also causes instrument dialogs to share the parent's taskbar entry. Is it a good thing? FamiTracker does that (at least on Wine), and it should make it easier to distinguish which of 2 instrument dialogs belongs to each of 2 BambooTracker main windows (which I've been confused by in the past).

(If I don't switch to QDialog, but give the instrument editors parents anyway, the dialog window contents are plastered within the main window.)

Tested on Linux. Will test on Windows at some point.

## Questions

I'm not convinced `std::shared_ptr<QWidget/QDialog>` is a good idea (the QDialogs survive being closed and are only destroyed when you delete the instrument or open a new document), but it works out in practice, so it's OK I guess?

Should `InstrumentEditorFMForm`/etc. inheriting from QDialog be renamed to include "Dialog" in the names?

### DIFF
--- a/BambooTracker/gui/instrument_editor/instrument_editor_adpcm_form.cpp
+++ b/BambooTracker/gui/instrument_editor/instrument_editor_adpcm_form.cpp
@@ -33,7 +33,7 @@
 #include "gui/gui_utils.hpp"
 
 InstrumentEditorADPCMForm::InstrumentEditorADPCMForm(int num, QWidget *parent) :
-	QWidget(parent),
+	QDialog(parent),
 	ui(new Ui::InstrumentEditorADPCMForm),
 	instNum_(num),
 	isIgnoreEvent_(false)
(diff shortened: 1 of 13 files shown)


nyanpasu64:
This matches Dn-FamiTracker behavior on Windows 10.

OPNA2608:
Seems okay to me.

> the QDialogs survive being closed and are only destroyed when you delete the instrument or open a new document

I checked with htop - creating 128 new instruments & opening their forms - and I see what you mean. If you can fix it, we'd appreciate a PR.

> Should `InstrumentEditorFMForm`/etc. inheriting from QDialog be renamed to include "Dialog" in the names?

The codebase it struggling abit with consistency in general, as you noticed awhile go. It's fine as-is for now.

Maybe we could track such problems in a meta issue to document & address them together?

PR #426, opened 2021-10-31
Update RtAudio submodule to upstream 5.2.0

nyanpasu64:
Fixes RtAudio randomly failing to discover PulseAudio devices, and *always* failing on pipewire-pulse (https://github.com/thestk/rtaudio/pull/307). So does it fix PipeWire? ...well kinda. The PipeWire side is still a broken mess.

In *both* PipeWire's ALSA and Pulse API modes, `RtAudio::stopStream()` (calling a "drain" function in both ALSA and Pulse) causes subsequent `RtAudio::startStream()` to never actually output audio. So BambooTracker works fine, until you press New (or unless you pass a module as a CLI argument) which stops and starts the stream, at which point the audio thread fills the buffers (which never get played) then hang forever.

I've proposed a fix to pipewire-alsa (a client-side library) at https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1467#note_1124276, but I haven't looked at pipewire-pulse (a daemon) to fix the bug there.


nyanpasu64:
you need to fix the ubuntu CI to 18.04 or 20.04 or another distro, github actions is no longer hosting 16.04 VMs: https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

nyanpasu64:
Rebased. Works on Linux with ALSA/Pulse talking to PipeWire. Didn't test on Windows, but there seems to be changes like `IAudioClient3` support (hopefully lower latency, unsure if BambooTracker actually benefits). Wish BambooTracker didn't suffer from linear interpolation, and low buffer lengths didn't result in detuning.

nyanpasu64:
Rebased again, updated RtAudio to 5.2.0 (https://github.com/thestk/rtaudio/releases/tag/5.2.0).

I've spent over a month building my own BambooTracker packages from a modified "build from local clone" PKGBUILD, and periodically rebasing my changes on upstream. `paru -Syu` either fails to pull upstream changes, or replaces my patched BambooTracker with upstream which can't output to pipewire-pulse. (I didn't try creating a PKGBUILD which checks out this branch and rebases on master, in an idempotent way that doesn't unnecessarily modify files and trigger rebuilds.)

----

I can't update RtAudio to master because they merged the no-exceptions branch (https://github.com/thestk/rtaudio/commit/0f4b52ceeef7a7da842d28990667972522f8c507) which breaks the API. If I checkout master, then BambooTracker no longer builds, since `AudioStreamRtAudio::initialize()` complains `RtAudioError` is not found.

Currently, when `audio->openStream()` fails, BambooTracker shows a dialog with contents taken from `RtAudioError::getMessage()`. After the merge, `RtAudioError` was removed, `audio->openStream()` only returns an enum code (no error message string), and `RtApi :: error()` only reports the message by calling the `std::function<...> RtApi::errorCallback_` callback. To my knowledge, you can't access the protected `RtApi::errorText_` field without patching RtAudio to expose it through a getter or similar.

If you upgrade to no-exceptions and want to keep printing error messages, (I think) you would have to set a callback. The issue is that `RtApi :: error()` (which calls your callback) gets called on both on the UI and audio threads, and your callback needs to act differently based on which thread it's running on.

One idea is to edit `AudioStreamRtAudio::initialize()`, replace the start of the `try` block with `audio->setErrorCallback(...);`, replace the end of the block with `audio->setErrorCallback({});` to reset it, and pray that the audio thread doesn't start and call `RtApi :: error()` before we call `setErrorCallback({})`. I'm not sure that's the case.

If I understand correctly, `RtApi :: openStream()` calls (eg.) `RtApiAlsa :: probeDeviceOpen()` which calls `pthread_create` (which as far as I can tell, begins execution immediately, rather than suspending the thread). In the case of ALSA, `alsaCallbackHandler` calls `RtApiAlsa :: callbackEvent()` which (in case of `STREAM_STOPPED`) waits for `AlsaHandle::runnable` to become true (it's initially false). I *think* RtAudio defaults to `STREAM_STOPPED`. As a result, `RtApiAlsa` can't continue executing and call `RtApi :: error()` before the stream is started.

I'm not sure if this will remain the case, and I didn't study every other backend to ensure they share this property (that if you call `RtApi::setErrorCallback({})` when `RtAudio :: openStream()` returns, then `RtApi :: error()` cannot possibly be called on the newly spawned audio thread). I *think* `RtApiPulse` is the same way.

If (in theory) the audio thread starts up, but immediately fails and calls `RtApi :: error()`, there's an endless zoo of race conditions between the GUI thread calling `RtApi::setErrorCallback({})` and resetting `errorCallback_`, the audio thread reading `errorCallback_` to check whether to call it (this is a data race), and the audio thread calling `errorCallback_` which tries to show a message box on the wrong thread (cue firework explosions).

One possible way out is to set a callback which checks the current thread ID and only opens a message box if running on the GUI thread.

----

How do you plan to handle errors during playback? Last time I checked, when you disable/unplug an audio device while BambooTracker is outputting to it through WASAPI, the program `std::terminate()`s from an uncaught exception on the audio thread, rather than showing a status bar message or a dialog saying the audio failed.

Ideally, BambooTracker on WASAPI would track the current default device when you plug or unplug devices, but I'm not sure if RtAudio supports that now (one reason I'm no longer confident I want to use it in future projects).

PR #429, opened 2021-11-07
Format keyboard keys

cxong:


OPNA2608:
Neat.

This list also exists in the wiki, could you update it there as well?
https://github.com/BambooTracker/BambooTracker/wiki/Shortcuts

cxong:
Done. Should we have one of the pages link to the other?

OPNA2608:
Wiki is much easier for people to extend & improve, the markdown file is what's getting packaged in a release. I tried to render the wiki for release documentation in the past but it got really hacky really quickly, so I gave up on that. It'll have to do like this...

PR #442, opened 2022-02-25
Don't focus main panel spinboxes on mouse scroll

### LINKED ISSUE #441: Scrolling song settings/step spinboxes should not defocus pattern editor
> #### Checklist
> 
> 
> 

---

## Bug Description

Some people have said they dislike how, when you scroll the mousewheel on the tempo/speed/step spinboxes, BambooTracker transfers focus from the pattern editor to the spinbox.

## How to reproduce

1. Open BambooTracker.
2. Scroll on the speed or step spinboxes.
3. Try to toggle edit mode or enter notes.

I have a custom spinbox which fixes this issue (https://gitlab.com/exotracker/exotracker-cpp/-/blob/b1d23ce36b204c9d54f74455b5bde02da4310a8a/src/gui/main_window.cpp#L264-284), but BambooTracker creates these spinboxes in https://github.com/BambooTracker/BambooTracker/blob/master/BambooTracker/gui/mainwindow.ui, and I'm unfamiliar with creating custom classes and adding them to .ui files.

## System Information

- **Operating System**: Linux, macOS, probably Windows
- **BambooTracker Version**: unstable cb59c69f1ed9359963ffb7b5203b5c58a62b0d3e
- **Build Type**: Built in Qt Creator

nyanpasu64:
Some people have said they dislike how, when you scroll the mousewheel on the tempo/speed/step spinboxes, BambooTracker transfers focus from the pattern editor to the spinbox. As a result, you can't toggle edit mode or enter notes, until you click on the pattern editor again.

This changes the spinboxes so they don't acquire focus when scrolled.

It would be even better if clicking the spinbox arrows didn't focus the spinbox instead of the pattern editor, but I couldn't figure out how to make this work. If I call `QSpinBox::setFocusPolicy(Qt::NoFocus)` and `lineEdit()->setFocusPolicy(Qt::StrongFocus)`, clicking the arrows (not the line edit) still focuses the line edit and unfocuses the pattern editor.

## Concerns

- I don't know why Qt Creator changed `<height>24</height>` in the .ui file. Is this a problem?
- Calling `QSpinBox::lineEdit()` may be too verbose.

## Preventing the instrument list from stealing pattern editor focus?

It would also be nice if clicking the instrument list didn't defocus the pattern editor. I tried editing `BambooTracker/gui/mainwindow.ui` and changing it to `Qt::TabFocus`, which almost works. But when you click "rename", the pattern editor gets unfocused anyway, and the instrument list gets a text field with a selection. However you can't click on the instrument list to end the rename (clicking the order or pattern editor works). And when you press Enter to close the rename text field, then the instrument widget is focused anyway (not the pattern editor).

I tried making`MainWindow::renameInstrument()` focus the instrument list before starting the rename, which didn't fix this. Is it too hacky to set `ui->instrumentList` to `StrongFocus` when a rename starts, and back to `TabFocus` when it ends?

Fixes #441.


rerrahkr:
Thank you for your request.
It seems to work well as far as I can see.
I haven't been able to confirm the behaviour of the instrument list you mentioned, but if it doesn't seem easy to implement your idea, it may be better to leave it the way it behaves now.

Sorry, the conflict was caused by handling another issue. I can fix it if you like, but would you like to do it yourself?

PR #450, opened 2022-05-25
Upload source tarballs on releases

### LINKED ISSUE #449: [Feature] - Create source tarball of releases containing all gitmodules
## Requested Feature

I know how to clone the git repository recursively in order to checkout all gitmodules but for packaging purposes (I am packaging/porting the FreeBSD version) that is not very convenient.

Could you please create source tarballs that contain all gitmodules? Before you mention the assets that allow to download a `.tar.gz`, those don't contain the gitmodules:

image

Providing a tarball that contains all the submodules has the following advantages for packager:

* tarball contains all submodules
* at the exact version it was released (not just a lose branch; yes submodules have the capability to use tags but this project doesn't do that)
* packaging frameworks can use simple, single fetch (fetch, wget) to get the entire source code

Thank you for your consideration.

OPNA2608:
Closes #449

Example run: https://github.com/OPNA2608/BambooTracker/releases/tag/v0.0.0

Bildschirmfoto von 2022-05-25 13-28-38
Bildschirmfoto von 2022-05-25 13-29-02

Does the tarball look okay to you @ehaupt?

### DIFF
--- a/.github/workflows/src-tarball.yml
+++ b/.github/workflows/src-tarball.yml
@@ -1,0 +1,52 @@
+name: Upload Source Tarball
+
+on:
+  push:
+    tags: 'v*.*.*'
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Identify build type.
+        id: identify-build
+        run: |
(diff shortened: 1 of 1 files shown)


ehaupt:
Looks good. It now contains all the submodules at the time of the release. Thank you @OPNA2608, that helps a lot.

PR #458, opened 2022-07-17
Update Nixpkgs version to 22.05

OPNA2608:
Should be updated before the next release. Will test before merging.


OPNA2608:
Seems to work fine, merging.

Issue #468, opened 2022-12-04
macOS - moving sliders in instrument editor moves other sliders (regression in version >0.5.0)


cxong:
> #### Checklist
> 
> 
> 
> - [x] I am reporting exactly 1 bug with this issue.
> - [x] This bug hasn't already been reported.
> - [ ] This bug hasn't already been fixed in the latest development build.

---

## Bug Description

In the instrument editor, when moving some sliders, the other sliders move as well, at different rates, when the value gets near 0.

## How to reproduce

1. Edit an instrument
2. Drag any slider so that it approaches 0
3. The other sliders will start to move

![slider](https://user-images.githubusercontent.com/1083215/205486255-74bb0830-3965-4861-b4bc-3dd246457de3.gif)

Bug can be also seen in official release 0.5.2, 0.5.3

## System Information

- **Operating System**: macOS Ventural 13.0
- **BambooTracker Version**: v0.5.1
- **Build Type**: Official

rerrahkr:
Thank you for your report. That is so terrible. I don't have a Mac so I can't check right away, but I'll check later to see if I changed anything before.

rerrahkr:
Does this glitch also affect the FM instrument or is it just a user interface issue?

cxong:
no - also if you look closer the numbers aren't changing, only the position of the sliders

rerrahkr:
It seems a bug caused by Qt: https://bugreports.qt.io/browse/QTBUG-98093.
In #470 I try to update Qt to the latest, but build is not working well. If that does not solve the problem, I need to create a custom slider to correct the drawing.

cxong:
Tested ^ under macOS ventura (13.0), lg2m

rerrahkr:
Updated to Qt6 in e9c5284, it would solve this issue.

PR #470, opened 2022-12-10
Use Qt6 for macOS build

rerrahkr:
#468

Older versions of Qt seem to have a problem with QSlider not rendering properly with the recent Monterey update (<https://bugreports.qt.io/browse/QTBUG-98093>).

This is a PR to see if changing the dev build in Appveyor to use Qt6 solves this problem.
In some cases, the GHA release build recipe may also need to be modified.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -58,6 +58,7 @@ environment:
       DEST: BambooTracker-v%APPVEYOR_BUILD_VERSION%-windows-xp-32bit.zip
       MAKE: mingw32-make
       RELEASE_BUILD: true
+    # macOS Catalina
     - APPVEYOR_JOB_NAME: for macOS
       APPVEYOR_BUILD_WORKER_IMAGE: macos-catalina
       PLATFORM: macos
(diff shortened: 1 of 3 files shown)


OPNA2608:
Times out because AppVeyor's macOS image is on 10.15 Catalina for which Homebrew doesn't carry a prebuilt version of Qt6, so it tries to build Qt6 from source:

image
image

Sticking with Homebrew's Qt will have to involve an upgrade to a newer macOS image.

Alternatively, we could try using one of the pre-installed Qt6's instead:

image

rerrahkr:
I missed the prebuild Qt. macOS jobs in GitHub Actions use Big Sur, so homebrew should work fine.

OPNA2608:
Feel free to take some of my changes from here if you'd like, replaces the AppVeyor Homebrew usage in favour of the pre-installed Qt 6.4 and no JACK support (saves time where it would otherwise have to update some Homebrew stuff, build Catalina packages that got removed etc): https://github.com/BambooTracker/BambooTracker/compare/master...OPNA2608:BambooTracker:appveyor_qt6_preinstalled

rerrahkr:
Mergerd. Thanks for your help!

PR #471, opened 2022-12-13
Add PMB bank import

OPNA2608:
This is a bank format used on the FM Towns. I still have many uncertainties about the bank metadata, but the sample data itself is easy to support.

The FM Towns isn't my particular forte but I know of 2 games & sound drivers that use PMB banks:
- Xenon (PMD): [xenon.pmb](https://github.com/BambooTracker/BambooTracker/files/10217389/xenon.pmb.zip)
- Desire ("KML/DOFMD"): [Desire's pmbs](https://github.com/BambooTracker/BambooTracker/files/10219945/Desire_pmbs.zip)

~~I still wish to check this against Desire's 3 PMB banks before finalising the PR, but I'm not on my PC with the required files & disc dumps right now.~~
Checked them now, work fine. I'll give the sample conversion another look cuz I just understood more about the format used, but should be ready otherwise.

### DIFF
--- a/BambooTracker/BambooTracker.pro
+++ b/BambooTracker/BambooTracker.pro
@@ -106,6 +106,7 @@ SOURCES += \
     io/io_utils.cpp \
     io/opni_io.cpp \
     io/p86_io.cpp \
+    io/pmb_io.cpp \
     io/ppc_io.cpp \
     io/pps_io.cpp \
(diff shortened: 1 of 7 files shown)


rerrahkr:
LGTM. Thank you for your contribution!

Issue #436, opened 2021-12-01
[Feature] - When exporting VGM, save mixing levels in extra header?


nyanpasu64:
> #### Checklist
> 
> 
> 
> - [x] I'm requesting exactly 1 feature with this issue.
> - [x] This feature hasn't already been requested.
> - [x] This feature hasn't already been implemented in the latest development build.

---

## Requested Feature

Currently when enabling mixing levels in BambooTracker (eg. making the SSG quieter than default), .vgm file exports don't contain mixing levels, leading to the SSG being louder in .vgm export than in the tracker.

The mixing levels could be saved in the .vgm file's extra header. VGM 1.70 (2014) added support for an extra header ([wiki](https://vgmrips.net/wiki/VGM_Specification#Extra_Header)), which can contain chip volume offsets. But BambooTracker currently doesn't write the [pointer to the extra header](https://vgmrips.net/wiki/VGM_Specification#:~:text=%20vgm%201.70%20additions%3A%20).

Is the extra header supported in players? ValleyBell [says](https://discord.com/channels/218044739893592064/755901971633799228/915385477266296862) libvgm supports the extra header, because "pretty much all PC-98 rips use it".

> https://vgmrips.net/forum/viewtopic.php?t=3747 example pack with SSG adjust

I don't think copying the offset directly from BambooTracker's internal mixer is good enough, since in_vgm 0.50.1's SSG is 5-6 dB louder than BambooTracker's SSG with no mixer active (using FM as a baseline, see below). You'll have to multiply the SSG volume by 0.3 or so, while converting module-specific mix levels to .vgm format.

@ValleyBell describes .vgm mixing as follows:

> The "absolute value 100%" = the default volume of the libvgm core (which can be really arbitrary)
> The "relative value 100%" = the default volume of the chip in VGMs (different from the default sound core volume)
> VGMs come with a pre-set balance for certain chips. e.g. YM2612's default VGM volume is 100%, SN76489's default volume is 50%, so that MegaDrive VGMs sound good
> In case of the YM2203/2608/2610 SSG, it is highly recommended to use a relative volume, as it changes the SSG volume based on the volume of the FM part.

is 100% relative volume equivalent to no extra header?

> yes

So in that case, it should be sufficient to output BambooTracker's FM and SSG volume levels, but multiply the SSG volume level by 0.3 or so.

## How much quieter should SSG be? (I don't know.)

[...]

PR #472, opened 2022-12-27
Add VGM export level configuration

rerrahkr:
Feature mentioned in #436.

The VGM export settings dialog now allows the user to select whether to write the setting of SSG volume level in the extra header.
Also, you can customize the volume level offset.

I don't know if my calculation is wrong, but it seems that if I set the offset to -1.0 dB, the volume will be the same as the BambooTracker's output. I'll check it out more...

### DIFF
--- a/BambooTracker/io/export_io.cpp
+++ b/BambooTracker/io/export_io.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2021 Rerrah
+ * Copyright (C) 2019-2023 Rerrah
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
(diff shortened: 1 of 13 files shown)


BambooTracker/io/export_io.cpp
```diff
@@ -107,14 +107,36 @@ void writeVgm(BinaryContainer& container, int target, const std::vector<uint8_t>
...
 
+	// 0xbc: Extra header offset
+	*reinterpret_cast<uint32_t*>(header + 0xbc) = mix ? (0x100 - 0xbc) : 0;
+
 	container.appendArray(header, 0x100);
 
+	// Extra header
+	if (mix) {
+		// Extra header size
+		container.appendUint32(17);
```
ValleyBell, line 118: This should actually be only "12".

I may need to clarify the specification: The "extra header size" covers only the pointer list itself and not the additional data that is pointed to.

rerrahkr:
It looks fine, so I'll merge it. Thanks for the help.

PR #494, opened 2023-07-31
Fix definition

rerrahkr:
Fix 832df71

### DIFF
--- a/submodules/RtMidi/RtMidi.pro
+++ b/submodules/RtMidi/RtMidi.pro
@@ -50,6 +50,7 @@ else {
 
   equals(pthread_required, "true") {
     LIBS += -lpthread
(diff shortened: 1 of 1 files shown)


OPNA2608:
This is an RtMidi bug, fixed by https://github.com/thestk/rtmidi/commit/f6e737827b85cee0cc20e553b8d21e20a2b899a7. Maybe bump past v5.0.0? Bugfix https://github.com/thestk/rtmidi/commit/9d863beb28f03ec53f3e4c22cc0d3c34a1e1789b also comes to mind, so maybe up to that?

rerrahkr:
Ah, it was dangerous to apply 5.0.0. Since there is no reason to update aggressively, I will revert to the previous version for the time being.

PR #495, opened 2023-08-12
Clarify detune effect description

cxong:
Original description mentioned effect range (00-FF) but did not say that 80 is no effect, this change clarifies what the ranges mean


OPNA2608 (commented):
IMO, this should mention that `80` resets it back to normal. Omitting `80` looks like it's an out-of-range / ignored value.

Precedent for out-of-range / ignored values going unmentioned: Brightness
https://github.com/BambooTracker/BambooTracker/blob/7b120d5cf34b1f6b1a16ba41a49f9d2351fb1e35/BambooTracker/gui/effect_description.cpp#L72
https://github.com/BambooTracker/BambooTracker/blob/7b120d5cf34b1f6b1a16ba41a49f9d2351fb1e35/BambooTracker/playback.cpp#L890-L894

Issue #506, opened 2024-04-29
[Windows|macOS|Linux|BSD] - Can FTBFS on Clang due to C++17


OPNA2608:
> #### Checklist
> 
> 
> 
> - [x] I am reporting exactly 1 bug with this issue.
> - [x] This bug hasn't already been reported.
> - [x] This bug hasn't already been fixed in the latest development build.

---

## Bug Description

Seeing this in Nixpkgs on the Qt6 macOS builds.

```
[...]

Issue #505, opened 2024-04-13
[Windows|macOS|Linux|BSD] - FTBFS on GCC 14


ycollet:
> #### Checklist
> 
> 
> 
> - [X] I am reporting exactly 1 bug with this issue.
> - [X] This bug hasn't already been reported.
> - [X] This bug hasn't already been fixed in the latest development build.

---

## Bug Description

I am trying to build Bambootracker with gcc-14.0.1 on Fedora 40.
The build fails with the following error:
```
gcc -c -pipe -std=gnu11 -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -Wall -Wextra -Wall -Wextra -Werror -pedantic -pedantic-errors -D_REENTRANT -fPIC -DQT_DEPRECATED_WARNINGS -DQT_NO_DEBUG -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -I. -Iinstrument -Imodule -I../submodules/emu2149/src -I../submodules/RtAudio/src -I../submodules/RtMidi/src -I/usr/include/qt5 -I/usr/include/qt5/QtWidgets -I/usr/include/qt5/QtGui -I/usr/include/qt5/QtCore -I. -I. -I/usr/lib64/qt5/mkspecs/linux-g++ -o wopn_file.o format/wopn_file.c
make[1]: Leaving directory '/builddir/build/BUILD/BambooTracker/BambooTracker'
format/wopn_file.c: In function â€˜WOPN_SaveBankToMemâ€™:
format/wopn_file.c:555:17: error: â€˜strncpyâ€™ output may be truncated copying 32 bytes from a string of length 32 [-Werror=stringop-truncation]
  555 |                 strncpy((char*)cursor, bankslots[i][j].bank_name, 32);
      |                 ^
cc1: all warnings being treated as errors
```

## System Information

[...]

PR #508, opened 2024-04-29
Make Werror optional for QMake

OPNA2608:
To opt out of it (i.e. due to deprecations), `CONFIG+=no_warnings_are_errors`.

To work around warnings like #505 and #506, the CMake build has a default-enabled `WARNINGS_ARE_ERRORS` option. It has been awhile since I really used QMake, so I don't remember if `CONFIG` options can be default-added without overriding command line arguments. So do the next best thing, and add an inverted `no_warnings_are_errors` `CONFIG` option to stop `-Werror -pedantic-errors` / `/WX` from being added.

### DIFF
--- a/qmake/variables.pri
+++ b/qmake/variables.pri
@@ -32,7 +32,10 @@ message("Qt is version" $$QT_VERSION)
 msvc|clang|if(gcc:!intel_icc) {
   msvc {
     message("Configured compiler is MSVC")
-    CPP_WARNING_FLAGS += /W3 /WX
+    CPP_WARNING_FLAGS += /W3
+    !no_warnings_are_errors {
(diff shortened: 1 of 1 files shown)


rerrahkr:
In #509 we changed to C++17 for building, is it necessary to merge this PR? #505 and #510 seem to be errors caused by  compiler not the language version.

OPNA2608:
We still need this, because #509 only switched to using C++17 on every compiler, but doesn't treat any of the associated deprecations. So an escape hatch like this is still useful.

The CMake script already has an option like this, so bringing the QMake one on par with it shouldn't be a problem.

rerrahkr:
Okay, I'll merge this. Thanks for your contirubution!

PR #509, opened 2024-05-04
Use C++17

rerrahkr:
#506 

Try to use C++17 to avoid errors during compilation.


OPNA2608:
Will apply this in Nixpkgs and see what it says there.

OPNA2608 (approved):
Works insofar that we're now requesting C++17 everywhere:
- `g++ -c -pipe -O2 -std=gnu++1z ...`
- `/Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -c -pipe -stdlib=libc++ -O2 -std=gnu++1z ...`

Although we have nothing on CI that's new/strict enough about the C++17 deprecation to actually give us a warning here, unlike what's on Nixpkgs for macOS:
- Our CI:
  ```
  Project MESSAGE: Qt is version 6.4.0
  Project MESSAGE: Configured compiler is Apple LLVM
  Project MESSAGE: Compiler is version 13.0.0
  ```
- Nixpkgs:
  ```
  Project MESSAGE: Qt is version 6.7.0
  Project MESSAGE: Configured compiler is LLVM
  Project MESSAGE: Compiler is version 16.0.6
  ```

But FWIW, the standard bump itself appears to work in all of our checked environments.

rerrahkr:
Thank you for checking this.

PR #514, opened 2024-07-06
Bump macOS 12

rerrahkr:
[GitHub runners for macOS 11 was deprecated](https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/#macos-11-deprecation-and-removal), so we need to update to 12.

At the same time, the image used by Appveyor is changed to 12. In addition, the version of Qt will be updated.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -58,9 +58,9 @@ environment:
     #   DEST: BambooTracker-v%APPVEYOR_BUILD_VERSION%-windows-xp-32bit.zip
     #   MAKE: mingw32-make
     #   RELEASE_BUILD: true
-    # macOS Big Sur
+    # macOS Monterey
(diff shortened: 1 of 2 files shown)


OPNA2608:
```
Error: No such keg: /usr/local/Cellar/python@3.10
Error: Process completed with exit code 1.
```

https://github.com/BambooTracker/BambooTracker/blob/edd1bd462f2eada101e83c97b2f77ba271dd4758/.github/workflows/macos.yml#L99-L102

Try removing this, will need to see if Homebrew complains about python mismatches again.

rerrahkr:
Encountered a sprintf error. A workaround was provided in #508, so I try to set `CONFIG+=no_warnings_are_errors` to qmake.

rerrahkr:
The error seems to have been resolved. Thanks for your help!

PR #540, opened 2026-02-08
ci: Switch nixpkgs channel to rolling, build Qt6 derivation, drop Linux release

OPNA2608:
[`bundle` is known-borked right now](https://github.com/nix-community/nix-bundle/issues/125). Tempted to just drop our Linux releases. Users would very likely be better off with either their distro's packaging of BambooTracker, or a binary that they built from source themselves.


OPNA2608:
I'm gonna go ahead with this. I don't really have the capacity to maintain a fully distro-agnostic Linux release, and I'm a sucker for "let distros run builds to get knowledge about potential issues with the code".

If you stumble upon this post-merge, and absolutely *NEED* a precompiled Linux binary of the latest versions… Open an issue and make your case for the best way to achieve it I guess. :s

PR #544, opened 2026-07-01
Allow copying in Jam Mode, allow order copying when opening context menu outside of selection

OPNA2608:
image

Confusion about this happened to at least one user.

Selection is fully enabled in Jam Mode, so I suppose there should be no reason to disallow copying, since it doesn't affect the module.

The context menu is created ad-hoc, while the `Edit` menu at the top updates based on signals. I figured saving the last `selected` signal's value to determine if copying should be allowed in the created context menu made the most sense.

---

image

No mentions of this yet, but it's a similar restriction.

There should be no real reason to require the user to right-click exactly into the order list selection for the copy option to be available. We have a signal indicating when the selection in the element changes, so just reuse that.


OPNA2608:
Should hopefully be fine…

PR #549, opened 2026-07-11
submodules/*/src: Fix compatibility with newer CMake

OPNA2608:
Should hopefully fix the modern Windows CI, which uses a recent-ish CMake version.

- emu2149:
  - Switch to BambooTracker/emu2149
  - Pull in fix for newer CMake
- RtMidi:
  - Update to 6.0.0


OPNA2608:
CMake 4 CI seems happy, sending it.

